PR #42, opened 2026-04-21
fix(console): per-request connection in SafeContext for multi-DB hosts

LeahArmstrong:
## The gap

`SafeContext` stored the database connection as `@connection` at initialization time. On hosts using ActiveRecord multi-DB (`connects_to`) or horizontal sharding, the rolled-back transaction was opened on that captured connection while queries could run on a different pool's connection — so the rollback did not cover writes on the shard.

## The fix

Remove `@connection` from `SafeContext`. Introduce a private `SingleConnectionPool` struct that wraps a bare connection object so the `connection:` kwarg follows the same code path as `pool:`. All execution now flows through `@pool.with_connection { |conn| run_with_timeout(conn) }` — a fresh connection is leased on every `#execute` call, not captured at init time.

`rack_middleware.rb` already used `pool: ActiveRecord::Base.connection_pool` (correct). `exe/woods-console` still uses `connection:` (bridge/stdio mode); it continues to work unchanged because `SingleConnectionPool` wraps it transparently.

## New specs

Added a `multi-DB / shard pool coverage` describe block to `safe_context_spec.rb`:
- Proves rollback fires on the shard pool connection when a shard pool is passed
- Asserts `@connection` ivar is absent in both `connection:` and `pool:` construction forms
- Verifies the `connection:` form still delivers a rolled-back transaction per call

## Test plan

- [x] `bundle exec rspec spec/console/safe_context_spec.rb` — 31 examples, 0 failures
- [x] `bundle exec rake spec` — 3759 examples, 0 failures, 2 pending (tiktoken not installed)
- [x] `bundle exec rubocop -a lib/woods/console/safe_context.rb spec/console/safe_context_spec.rb` — no offenses

### DIFF
--- a/lib/woods/console/safe_context.rb
+++ b/lib/woods/console/safe_context.rb
@@ -19,15 +19,22 @@ module Woods
     #
     # Two construction modes are supported:
     #
-    # - `connection:` — the legacy single-connection form. Useful in tests
-    #   and for callers that already manage their own connection lifecycle.
-    # - `pool:` — the per-request form. Each call to {#execute} leases a
-    #   fresh connection via `pool.with_connection { |c| ... }`, so the
-    #   connection is returned to the pool immediately after the block. The
-    #   leased connection is also exposed via
-    #   `Thread.current[:woods_console_leased_connection]` so dispatch
-    #   handlers (e.g. EmbeddedExecutor#active_connection) can reuse the
-    #   same connection without re-leasing.
+    # - `connection:` — wraps the supplied connection in a single-use pool
+    #   adapter so the execution path is identical to the `pool:` form.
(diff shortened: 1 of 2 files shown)


LeahArmstrong:
Added a behavioral regression guard in `332107b` per review feedback.

The shard-pool specs proved "new code reaches the shard," but the regression signal on the legacy `connection:` form was only structural (`@connection` ivar absence check). Added a spec that stubs `pool.with_connection` with a call-count passthrough and asserts all 3 execute calls route through it — would fail if someone re-introduced captured-connection fallback even with the ivar check still passing.

32/0 specs in `safe_context_spec.rb`.

PR #79, opened 2026-04-23
fix(mcp): Bootstrapper hydrates from Snapshotter dumps at boot

LeahArmstrong:
The last-mile fix that closes the Shape-2 persistence arc started in PR #73.

## What was broken

PR #77 shipped the real `Snapshotter::Vector` and `Snapshotter::Metadata` and wired the **dump side** — end-of-embed writes `vectors.bin` + `metadata.msgpack` + `woods.json` into `dumps/<ISO8601>/` and flips the `latest` pointer. But the **read side** was still running PR #75's stub:

```ruby
def self.build_retriever_from_config(config, _resolved, _artifact)
  # Snapshotter hydration is stubbed in PR 2 — the Builder constructs
  # empty InMemory stores, same as before. PR 3 wires in the real
  # Snapshotter.load_or_empty call after build so the returned
  # retriever sees hydrated vectors.
  Woods::Builder.new(config).build_retriever
end
```

PR 3 didn't wire it. Surfaced while re-running the plan audit before tackling follow-ups.

**Net effect on admin:** embed writes dumps, `woods-mcp` restarts, sees empty stores, semantic search returns nothing. The whole arc went unrealized until this fix.

## What lands

- **`Bootstrapper.build_retriever_from_config`** now calls `Snapshotter::Vector.load_or_empty` + `Snapshotter::Metadata.load_or_empty` for `:in_memory` stores, threading the hydrated stores into `Builder` via new `vector_store:` / `metadata_store:` kwargs.
- **`Builder.build_retriever`** accepts the optional stores so the Bootstrapper can inject hydrated ones. Builder still constructs fresh empties for callers that don't inject — zero change for existing code paths.
- **Durable backends** (pgvector, Qdrant, SQLite) never enter the Snapshotter path — the condition gates on `config.{vector,metadata}_store == :in_memory`.
- **Config-invalid is loud, file errors are soft.** `Woods::MCP::BootstrapError` subclasses (`DimensionMismatch`, `UnsupportedArtifact`, `ConfigMismatch`) re-raise at boot — we don't quietly ignore mismatches. Generic `StandardError` during read (file corruption, IO error) falls through to "start with empty store + warn" — same behavior as missing-dump.
- **`Builder.build_embedding_provider`** filters `SNAPSHOT_ONLY_KEYS` (currently just `:dimension`) from `embedding_options` before splatting into the provider constructor. Dimension lives in `embedding_options` for `ResolvedConfig`'s sake but isn't part of the Ollama/OpenAI API — without the filter, any config that declared a dimension would `ArgumentError` at provider construction. Real production bug hiding in the corner.

## Spec

New scenario in `spec/mcp/bootstrapper_spec.rb`:

1. Populate a `VectorStore::InMemory` + `MetadataStore::InMemory` in-process
2. Dump via the Snapshotter (mirroring what the Indexer does at end-of-embed)
3. `artifact.promote(dump_dir)` — flip the latest pointer
4. Call `Bootstrapper.build_retriever(index_dir: dir)` (fresh config, no in-process state)
5. Assert the returned retriever's internal stores have the seeded entries

This is the only spec that actually proves the Shape-2 boot path works — everything upstream (the round-trip integration spec, the Snapshotter specs) would have passed even with the stub still in place.

## Test plan

- [x] `bundle exec rake spec` — **4315 examples, 0 failures, 2 pending**
- [x] `bundle exec rubocop` — 424 files, 0 offenses
- [x] The new scenario covers the critical path missing since PR 3

## Why this wasn't caught earlier

The plan audit between PR 3 and PR 4 missed this because:
1. PR 3's spec coverage focused on the Snapshotter's own round-trip (works in isolation).
2. PR 3's Indexer spec used doubles for the Snapshotter (no real write/read cycle).
3. The integration spec in PR 3 invoked `Snapshotter.load_or_empty` directly, not through `Bootstrapper.build_retriever`.
4. The stub method's own comment said "PR 3 wires in the real call" — which turned out to be an aspiration, not a commitment.

Takeaway: specs that exercise the full boot path (not just the adapters in isolation) are load-bearing for multi-component arcs. The new bootstrapper spec is that coverage.


LeahArmstrong:
## Amendment: review-caught critical bugs + deep spec expansion

A post-PR audit with 4 parallel reviewers (goal verification, PR deep-dive, CHANGELOG gap, hidden-issue hunt) found **3 CRITICAL bugs this PR as originally-scoped would have shipped with**, plus several hardening gaps. All fixed in a single amended commit; PR scope is now the full "make Shape 2 actually work" payoff, not just the one hydration wiring fix.

### Critical bugs found + fixed

1. **Vector metadata silently discarded on hydration.** `Snapshotter::Vector#hydrate_store` hardcoded `metadata: {}` per rehydrated entry — the WVF1 binary format only carries the float blob (by design, for mmap). Per-vector metadata lives in `metadata.msgpack`. Consequence: type-filtered semantic searches silently returned zero results after any dump/reload. Fix: `Bootstrapper.populate_vector_metadata` back-fills from the metadata store after load.

2. **`Tasks.build_embed_indexer` never wrote `woods.json`.** The rake path didn't pass `resolved_config:`, `metadata_store:`, or `dump_retention_count:` to `Indexer.new`. `persist_snapshot` correctly skipped `write_config` when `@resolved_config` was nil — so `vectors.bin` + `latest` pointer got written, but `woods.json` didn't. The standalone MCP server then raised `MissingArtifact` at boot with no recovery. **The entire `:shared_filesystem` preset was broken.** Fix: thread all three kwargs through.

3. **OpenAI `woods.json` boot crashed with raw `ArgumentError`.** `ConfigResolver.populate_from_stored` didn't read `OPENAI_API_KEY` from env (by design, woods.json omits credentials). `OpenAI.new(**opts)` then raised `ArgumentError: missing keyword: api_key`, which isn't a `BootstrapError`, so `exe/woods-mcp`'s top-level rescue didn't catch it. Fix: read `OPENAI_API_KEY` from env, raise typed `MissingCredential` with actionable message when absent.

### Additional hardening

- Truncated `vectors.bin` raises typed `UnsupportedArtifact` not `NoMethodError`.
- `hydrated_vector_store` / `hydrated_metadata_store` propagate `ArgumentError` (config-bug) instead of swallowing as I/O failure.
- `ResolvedConfig.validate_schema_version!` uses `<=` (forwards-compatible) to match the binary snapshotters.

### Spec expansion (~1700 LoC, ~200 new examples)

- **`spec/integration/shape2_end_to_end_spec.rb`** (NEW) — 12 examples covering the actual end-to-end boot path, including the three critical bugs above.
- **`spec/mcp/bootstrapper_spec.rb`** — failure-mode blocks for soft-fail rescue + typed-error propagation.
- **`spec/storage/snapshotter/vector_spec.rb`** — failure-mode blocks for every header corruption class.
- **`spec/embedding/indexer_spec.rb`** — `dump_retention_count` configuration variants.

### CHANGELOG

Added comprehensive `[Unreleased]` entries for the full persistence arc (PRs #73–#79) — Added (12), Changed (7), Fixed (6) — grouped under a "Persistence & Bootstrap arc" heading to distinguish from the Ollama-tokenizer arc that preceded it.

### Full suite

- 4374 examples, 0 failures, 2 pending (unchanged `tiktoken_ruby` pendings)
- Rubocop clean on 425 files

### Lesson

The pre-amendment PR would have shipped with type-filtered search broken in production despite passing every spec. The audit caught it because the spec suite previously tested adapters in isolation — no spec went through `Bootstrapper.build_retriever(index_dir:) → dump → search with filter`. That's now the capstone test in `shape2_end_to_end_spec.rb`.

PR #116, opened 2026-04-28
Liam ERD integration v2 (rebased on main)

LeahArmstrong:
Tree-replay of the `erd` branch (88 commits) onto current main as a single commit so the integrated work is testable against latest main without disturbing the original `erd` branch (still at `f64cc67`).

## What's in here

- **Phase 1** — vendored Liam ERD assets, Rack middleware + Railtie mount, `erd_enabled` / `erd_path` config, model-to-schema generator
- **Phase 2** — `WoodsNode` for controllers/jobs/services/mailers, `erd_layers` config, toolbar layer toggles, sidebar focus mode
- **Phase 3** — perf (lazy layer creation, member-cap expand/collapse), multi-node focus + `?focus=` URL shareability, command palette with node search
- **Journey mode** — entry-point indexer + BFS walker, JourneyContext, banner/legend overlays, "Start journey from here" context menu, navigation-edges toolbar toggle

## Why squashed

Rebasing 88 commits onto main hit conflicts on `lib/woods.rb`, `spec/configuration_spec.rb`, and a few others where both branches added different config options to the same lists. For a testing branch the cumulative tree is what matters; per-commit history is preserved on the original `erd` branch.

The tree of this commit is bit-identical to `erd^{tree}` — verified before push.

## Test plan

- [ ] `bundle exec rake spec` on `erd-v2`
- [ ] Boot a `woods-testbed` variant against this branch, hit `/woods/erd`, smoke-test Phase 1/2/3 + journey mode


LeahArmstrong:
Closing — the squash approach silently dropped 105 files that landed on main since the erd branch's last sync (bridge protocol, credential scanning, recent docs, etc.). Recreating via a real merge so main's recent work is preserved.

PR #128, opened 2026-06-10
feat(unblocked): incremental sync via content-hash manifest

LeahArmstrong:
## Why

The Unblocked sync re-pushed the **entire** collection (~800–1200 `put_document` calls) on every run, ignoring extraction's incremental mode and burning most of the 1000-call/day API budget — capping deploys at ~67/day. This makes the sync incremental so it scales to per-deploy CI.

## What

A **`SyncManifest`** records the content hash + remote `document_id` of everything last pushed (JSON under the index dir, persisted across CI via the provider's cache). Each run now:

- **skips** documents whose built body hash is unchanged,
- **pushes** only new/changed documents,
- **deletes** documents whose source unit has disappeared.

Documents are upserted by URI, so a missing manifest (first run / CI cache miss) degrades to a correct full re-push that rebuilds the manifest. Steady state on an unchanged codebase costs ~0 calls.

## Key decisions

- **Hash the built document (title + body), not `source_hash`.** Bodies derive from metadata + dependents; `source_hash` (source code only) would miss dependent-driven changes.
- **Determinism precondition:** `DocumentBuilder` sorts every rendered collection so an unchanged unit always hashes identically. Covered by an order-independence spec.
- **Reconcile, not query, for change detection:** the live `GET /documents` returns no body, so it can't supply our hash. `Client#all_documents` (paginated, client-side collection filter) recovers `document_id`s on a manifest-less run so orphans can still be purged.
- **Mass-deletion guard:** refuse to purge >30% of a ≥10-doc manifest unless `UNBLOCKED_FORCE_PURGE=1` — protects against running against a partial index. Purge is also skipped when the daily budget is exhausted mid-run.

## Hardening (post-review commit)

A full 7-angle review of the branch surfaced 7 verified findings, all fixed in `fix(unblocked): harden sync from code-review findings`:

1. Public `sync_type`/`sync_type_partial` crashed when called without `sync_all` (nil state) — initialized in constructor.
2. `ensure`-block manifest save could fail an otherwise-successful sync and mask in-flight errors — now best-effort + warning.
3. Pagination: `after` cursor URL-encoded; nil-cursor full page no longer loops forever against the budget.
4. `UNBLOCKED_FORCE_FULL_SYNC=false` no longer *enables* the flag — truthy-set parsing (`1`/`true`/`yes`).
5. New `ApiError < Woods::Error` with HTTP `status`; purge treats delete-404 as already-gone instead of retrying forever.
6. Stale entries with nil `document_id` are resolved via one bounded `all_documents` sweep instead of being permanently undeletable.
7. `list_collections` bare-array `TypeError` fixed (live API shape).

## Branding

Adds the Tree Rings logo set under `assets/` (outlined-text wordmark SVGs + PNG exports with/without backgrounds); README uses the white-on-black wordmark PNG. The square mark doubles as the **default Unblocked collection icon** — `Client#create_collection` now falls back to `DEFAULT_ICON_URL`, since the live API **requires `iconUrl`** to create a collection (despite docs marking it optional; documented as a known quirk).

## Escape hatches

- `UNBLOCKED_FORCE_FULL_SYNC=1` — re-push everything (e.g. after a builder format change).
- `UNBLOCKED_FORCE_PURGE=1` — bypass the mass-deletion guard.

## Verification

- **Full suite green:** 4949 examples, 0 failures, 4 pending (optional-gem skips). RuboCop clean.
- **Live smoke (twice, incl. post-fixes)** against a live Team API token: create collection → `put_document` → `all_documents` → `delete_document` → delete collection. Self-cleaning, `SMOKE: PASS`.
- Ruby 4.0 build compat in a separate commit (`tokenizers` gated via `install_if`; `benchmark` declared). Lockfile unchanged.

## Release-readiness round 2 (team review of 5 specialist agents)

A five-agent review (code, tests, silent failures, comment accuracy, type design) drove a second hardening commit:

- **CI-visible failures:** the rake task exits non-zero on errors (dead-token runs can no longer stay green); budget exhaustion with partial progress is the tolerated cold-start shape.
- **CI doc fixes:** the GHA cache example now uses unique key + `restore-keys` (immutable cache entries froze the manifest at run 1); the Buildkite example uses the cache plugin (plain `artifact download` is build-scoped and never restored anything).
- **Standalone `require 'woods/unblocked/client'` fixed** (load-time `Woods::Error` resolution).
- **API shape locked pre-release:** `ApiError` requires `status:`; new `BudgetExhaustedError` replaces message-string matching; `SyncManifest` validates its schema version on load and warns on every discard.
- **Operational visibility:** purge delete failures surface in `stats[:errors]`; reconcile aborts on 401/403; an empty scrubbed body skips the push instead of overwriting the remote doc.
- **New tests:** two-run end-to-end with a real manifest, partial top-N fallout protection, guard boundary cases, ensure-save on the exception path, doc-id fallback, and a first `RateLimiter` spec.

## Open questions for the consuming pipeline

1. **Manifest persistence:** S3 cache plugin or artifact-API restore for the manifest? Upload is free if `artifact_paths: tmp/woods/**/*` already exists. Key on `UNBLOCKED_COLLECTION_ID`.
2. **Concurrency:** the gem does not lock — the sync step needs a `concurrency_group` for per-deploy cadence.
3. **Cold-start budget:** a cache-miss run on a ~1000-doc collection needs ~1005 calls vs the 1000/day default — converges over two runs unless `UNBLOCKED_DAILY_BUDGET` is raised.
4. **Cadence:** steady-state per-merge syncs cost ~0 puts — nightly schedules can be revisited.
5. **Collection setup:** one-time `create_collection` now defaults the icon; just supply name/description.


LeahArmstrong:
Folded in both observations from the consuming-pipeline review:

1. **Units without a `file_path` are now skipped** rather than pushed to the shared repo-root URI. This was worse than manifest churn — URIs are the upsert key, so any two such units were already overwriting each other's remote document. Skipped units are not tracked as current, so a stale repo-root doc from before the guard purges naturally. Two new specs.
2. **Docs note added**: the mass-deletion guard also fires on intentional large removals (type dropped from the sync set, `repo_url` change, big deletions) — re-run once with `UNBLOCKED_FORCE_PURGE=1` when expected.

Suite: 4968 examples, 0 failures. RuboCop clean.

PR #146, opened 2026-07-01
Query-driven Svelte Flow visualization

LeahArmstrong:
## Summary

Refactors the Svelte Flow visualization from "render the whole graph" to "render the result of a query." An agent can query Woods (`dependents`, `trace_flow`, `search`, …), hand the user an explorable map of exactly the relevant units, let them click a node to read its source and jump to the file, and share the result by URL or as a self-contained file.

Also fixes a latent bug that dropped **all** edges from the dependency-graph view, and clears the pre-existing rubocop debt across `lib/woods/svelte_flow/`.

## Motivation

The visualization generated a full mapping by default — a wall of hundreds/thousands of nodes the user had to prune by hand. The workflow we actually want starts from an answer the agent already has. This makes the query→render→explore loop the primary path, while keeping the full-graph paths intact.

## Changes

**Render a query result (Phase 1)**
- `GET /api/subgraph?nodes=A,B,C&depth=N&via=...` returns the induced subgraph over an identifier set, expanded by `depth` BFS hops and optionally relationship-filtered; `requested`/`dropped` report resolution.
- `?nodes=` URL param loads just that scoped set (`App.svelte` / `api.js`).

**Inspect a unit (Phase 2)**
- `GET /api/unit/:id/source` (identifier-keyed — no arbitrary-file-read surface) feeds a source pane in the detail panel with a `vscode://` editor link and, when `svelte_flow_repo_url` is set, a GitHub blob link pinned to the extraction's git SHA. Connected-unit references are highlighted in the source.
- New config `svelte_flow_repo_url` (env `WOODS_SVELTE_FLOW_REPO_URL`).

**Self-contained export (Phase 3)**
- `rake woods:map NODES=A,B,C [DEPTH=n] [VIA=...]` writes one self-contained HTML (scoped graph + sources inlined) that opens over `file://` with no server.
- Extracted `SubgraphScoper` and `SourceLinks` — shared by the live endpoints and the offline export so they can't drift; the middleware shed ~230 lines.

**Polish (Phase 4)**
- `@xyflow/svelte ^1.0 → ^1.6` (1.6.1): improved `fitView`, keyboard nav, a11y.
- `:via` edge coloring + on-canvas legend; depth-pulled neighbors dimmed so the queried set stands out.

**Baseline (Phase 0)**
- Fixed the root cause of the failing viz specs: consumers iterated `{ target:, via: }` edges as bare strings, so no edges rendered. Added the `EdgeData` normalizer and threaded `:via` through. Cleared all pre-existing rubocop offenses in `svelte_flow/`.

Design doc: `docs/SVELTE_FLOW_QUERY_REFACTOR.md`. Deferred (needs a product decision): exact line-span highlighting — the substring approximation ships now.

## Testing

- **Gem specs:** full suite green — `5183 examples, 0 failures, 2 pending` (run under `LANG=C.UTF-8`). 21 new specs across `SubgraphScoper`, `SourceLinks`, `StandaloneRenderer`, the exporter, and the two new middleware endpoints.
- **Lint:** `rubocop` clean (502 files, 0 offenses).
- **Browser (headless Chromium):** validated each phase end-to-end — server mode (`?nodes=`, source pane + links + highlighting, `:via` colors, legend, neighbor dimming) and the standalone file over `file://` (renders offline, source pane works, no JS/console errors).

## Checklist

- [x] Tests added/updated
- [x] `bundle exec rake spec` passes (via `bin/rspec`; `LANG=C.UTF-8` for the one locale-sensitive MCP spec)
- [x] `bundle exec rubocop` passes (via `bin/rubocop`)
- [x] CHANGELOG.md updated
- [x] Documentation updated (`docs/SVELTE_FLOW_QUERY_REFACTOR.md`, `docs/SVELTE_FLOW_VISUALIZATION.md`, `docs/CONFIGURATION_REFERENCE.md`)

---
_Generated by [Claude Code](https://claude.ai/code/session_01ERMDHn4auzeKqb6vWEza4X)_


LeahArmstrong:
Tested this branch end to end in a real Rails app (a large production admin, ~6.8k units). Both paths work: `rake woods:map NODES=…` produces a self-contained HTML that opens over `file://` with sources inlined, and server mode (`svelte_flow_enabled`, `/woods/visualize?nodes=…&depth=…`) renders live with the detail panel fetching source on demand. The query-scoping, `?center=`, and the on-demand `/api/unit/:id/source` all behaved correctly. Nice work.

Two categories of feedback below: one product gap on the MCP surface, and three rendering issues that all trace back to the edge/node data the frontend emits. Evidence is from the live `/api/subgraph` payload for a real model.

---

## 1. The MCP can't hand back a URL or a file — both need a shell

The visualization is only reachable two ways today, and neither is available to an MCP client:

- **No shareable URL.** The MCP tools (`lookup`, `dependents`, `search`, …) return graph data only. Constructing `/woods/visualize?nodes=…` is left to the caller, and the MCP has no way to know the host/port the Rails app is served at, so it can't produce a browsable link even in principle.
- **No file generation.** Producing the self-contained HTML is a rake task (`woods:map`), so an agent has to shell out to run it, then copy the file off the container. The MCP surface never touches the `svelte_flow` exporter.

The result is that the natural agent loop ("show me X") always drops out of MCP into shell + file plumbing.

### Proposed solution

Add a `visualize` MCP tool that reuses the code already on this branch:

- **Always:** call `StandaloneRenderer` in-process, write the scoped HTML, and return its path. This is a pure artifact write; no server needed. It closes the "generate a file without server commands" gap entirely.
- **Optionally:** when a base URL is configured (new `svelte_flow_base_url`, same spirit as the `svelte_flow_repo_url` you added), also return the `/woods/visualize?nodes=…&depth=…&via=…` link. Gate it on the config so the tool never fabricates a host it can't verify.

Signature could mirror the rake params: `visualize(nodes:, depth: 0, via: nil)` -> `{ file_path:, url: (if configured) }`.

---

## 2. Rendering issues (all in the edge/node emit layer)

These share one root: the subgraph payload underspecifies the edges, so the frontend can't draw them accurately. Sample edges from `/api/subgraph?nodes=ShippingProfile::RateCategory&depth=1`:

```json
{"id":"e-…RateCategory-Account","source":"ShippingProfile::RateCategory","target":"Account","type":"default","data":{"relationship":"dependency","via":"belongs_to","isCycle":true},"animated":true}
{"id":"e-…RateCategory-…Method","source":"ShippingProfile::RateCategory","target":"::ShippingProfile::Method","type":"default","data":{"relationship":"dependency","via":"has_many","isCycle":false}}
```

Nodes come back with `position: {x:0, y:0}` (layout is computed client-side).

### 2a. Edges don't anchor to the right place

Edges carry only `source`/`target` (node ids), never `sourceHandle`/`targetHandle`. With no handles, Svelte Flow attaches every edge to the node's default anchor, so an FK relationship like `Method.rate_category_id -> RateCategory.id` lands on the card border rather than on the two columns it actually connects. Visually the lines look like they join arbitrary points.

**Fix:** give each rendered column row a handle (`id="col-<name>"`) in the node component, thread the association's `foreign_key` (and target PK) into the edge data in the scoper/edge builder, and set `sourceHandle`/`targetHandle` on association edges. Non-column edges (include/extend/code_reference) can keep the default anchor.

### 2b. Paths overlap

Every edge is `type: "default"` (a bezier), and with the client-side layout the nodes end up stacked in a column, so parallel edges run over each other and through intervening cards. It reads as a tangle even at 17 nodes.

**Options:**
- Switch association edges to `smoothstep` (orthogonal) and run a real layout pass (dagre or elkjs) instead of relying on the current placement. This is the durable fix and pairs naturally with the handles from 2a.
- Lighter touch: keep bezier but add per-edge source/target offsets for parallel edges plus z-index so they separate. Reduces overlap without a layout dependency.

### 2c. `belongs_to` / `has_many` / `has_one` never render

`data.via` already carries the exact relationship for every edge (`belongs_to`, `has_many`, `has_one`, `include`, `extend`, `code_reference`, `factory_for`, `test_coverage`), but nothing renders it. The only edge distinction on canvas is the two-category legend (Association vs Reference) via color. The specific cardinality, which is the most useful thing about a model graph, is dropped on the floor.

**Options:**
- Custom edge type with a small label pill showing `via` (`belongs_to`, `has_many`, …). Lowest effort, reads clearly.
- Cardinality markers: `markerEnd` crow's-foot style driven by `via` (one-end for `belongs_to`/`has_one`, many-end for `has_many`). More work, but it turns the graph into a proper ER diagram.
- At minimum, expand the legend and per-edge styling beyond the two categories so the `via` data is at least distinguishable.

---

All three of 2a/2b/2c live in the same subsystem (the Svelte edge/node components plus the scoper's edge emit), so they're likely one focused change rather than three. Happy to open a follow-up issue with these broken out if that's easier to track than a PR thread.

LeahArmstrong:
Follow-up on the source detail panel specifically. Three issues, all verified against the running server.

### 1. It serves the extraction snapshot, not the live file

The source endpoint returns the stored copy, not the file on disk. `RackMiddleware` does:

```ruby
'sourceCode' => unit['source_code'] || unit[:source_code]
```

Two consequences:

- **It can drift from the working tree.** The panel shows whatever the source was at extraction time, so after any edit the preview is stale until the next `woods:extract`.
- **For models it duplicates the panel above it.** The stored source leads with the `annotate` schema block (`# == Schema Information / # Table: accounts / # Columns: …`), which is the same information already rendered in the COLUMNS section right above. So the first screenful of "source" is a restatement of the columns, not the actual class.

The intent (per the panel's purpose) is to show the real application file. In server mode you can read it live, keyed off the unit's own `filePath`, which keeps your no-arbitrary-file-read guarantee intact (the path comes from the resolved identifier, never from the client):

- **Server mode:** `File.read(unit.filePath)` when the file exists, fall back to the stored `source_code` otherwise (and always fall back for the offline export, which has no filesystem).
- **Optional:** strip or collapse the leading `annotate` block so the visible content is the class body rather than a duplicate of COLUMNS.

### 2. The panel is too small to read code in

The source pane is a short box pinned to the bottom of the detail sidebar. Even a small model doesn't fit, and Account (1000+ lines) is unusable in it. Options:

- Make the pane resizable or full-height within the sidebar.
- Or open source in a modal / dedicated panel with real height.
- Either way, add syntax highlighting and keep the connected-unit reference highlighting you already have.

### 3. The editor link can't resolve

The `vscode://` link is built client-side by prefixing `vscode://file/` onto the repo-relative path, so it comes out as `vscode://file/app/models/account.rb`, which points at a nonexistent `/app/models/account.rb`. There is no config for the local project root, so the link can never resolve to where the file actually lives on the reader's machine (for us, extraction runs in a container at `/app` while the editor is on the host at a completely different path).

**Fix:** add an editor-root config (a local absolute path), the editor-side analogue of the `svelte_flow_repo_url` you added for GitHub, and build the link as `vscode://file/<editor_root>/<repo-relative path>`. Building it in `SourceLinks` alongside `github_blob_url` (rather than in the frontend) would keep both links constructed in one tested place and let the offline export emit a correct editor link too.

---

1 and 3 are the same shape as the GitHub-link work already on this branch: the data/path is right, but there's no host-side mapping to turn a container/extraction path into something the reader's machine can open. A single "how do local paths map to this environment" config could feed both the source read and the editor link.

LeahArmstrong:
All feedback addressed in 3d6a8cd. Validated against a fixture in headless Chromium (server mode + `file://`); happy to have you re-run against the real admin.

## 1. MCP `visualize` tool — done

New always-on index-server tool: `visualize(nodes:, depth: 0, via: nil)`. It calls the standalone exporter in-process and always returns `file_path`; when the new `svelte_flow_base_url` config (env `WOODS_SVELTE_FLOW_BASE_URL`) is set it also returns the `/woods/visualize?nodes=…&depth=…&via=…` URL — never fabricated otherwise, exactly as you proposed. One extra: if the index dir is a read-only mount (host-side MCP reading a container volume), the file falls back to the system temp dir instead of erroring.

## 2a. Edges now anchor to columns

Root cause confirmed: the full-graph path (`/api/graph`) ran `AssociationEdgeBuilder` but the subgraph/neighbors path only ran the generic `EdgeBuilder`, so scoped payloads lost `type: "association"`, handles, and FK data. `SubgraphScoper` now emits model↔model relationships as association edges (`sourceHandle: "Order-account_id"`, `targetHandle: "Account-id"`, macro in `data.via`), deduped against the generic edges **in both directions** — with one deliberate improvement over the full-graph path: model→model pairs with *no* association metadata (plain code references) still render as generic edges rather than vanishing.

## 2b. Parallel edges separate

Took your lighter-touch option: edges stay smoothstep, but each edge on a side of the center gets a progressively larger `pathOptions.offset`, so the vertical runs spread across the column gap instead of stacking. (A real dagre/elk layout pass remains a candidate if the offsets prove insufficient at your scale — say the word and I'll take it up as a follow-up.)

## 2c. Relationships render

Two of your three options landed: association edges now carry a **label pill** (`belongs_to`, `has_many`, …), and the **cardinality markers** (crow's-foot / bar) work — they were silently broken everywhere by marker ids double-wrapped into `url('#url(#marker-bar)')`; fixed while in there. Color-by-category + legend stays for non-association edges.

Follow-ups I'd suggest tracking as issues rather than growing this PR: a proper layout engine (2b), and showing edges between two non-center queried nodes (the canvas currently only draws center-connected edges — predates this PR but matters more in query mode).

---

LeahArmstrong:
Source panel feedback also addressed in 3d6a8cd:

## 1. Live file, not snapshot — done

New shared `UnitSource` module (used by both the middleware endpoint and the standalone export): reads the file at the unit's **own extracted** `file_path` when it exists — the path never comes from the client, so the no-arbitrary-read guarantee holds — and falls back to the stored snapshot otherwise (the response carries `live: true/false`, and the panel shows an "Extraction snapshot" notice on fallback). The leading `# == Schema Information` annotate block is stripped in both modes (only when it sits in the file's leading comment region — magic comments above it survive, and a block later in the file is left alone).

## 2. Panel is now readable — done

While source is open the detail panel widens to `min(560px, 65vw)` and the code flows at full height inside the panel's own scroll (no more nested scrollbox). Added Ruby syntax highlighting via a small self-contained tokenizer (comments/strings/symbols/keywords/constants/ivars/numbers — no CDN, the standalone export stays offline-safe), composed with the existing connected-unit `<mark>`s.

## 3. Editor link resolves — done

Exactly the shape you suggested: new `svelte_flow_editor_root` config (env `WOODS_SVELTE_FLOW_EDITOR_ROOT`), and the link is now built server-side in `SourceLinks.editor_url` next to `github_blob_url` — one tested place, and the standalone export emits it too. With the root set you get `vscode://file/<root>/<repo-relative>`; without it, the stored absolute path (right when extraction and editor share a filesystem); a bare relative path yields **no** link rather than one that can't resolve.

Your "single path-mapping config" observation is right — `editor_root` is that mapping for the editor side, and `UnitSource`'s live-read covers the server side. If a third consumer shows up we can consolidate them behind one path-mapper.

---

LeahArmstrong:
Looked at [andrew2net/rails-schema](https://github.com/andrew2net/rails-schema) (thanks for the pointer) and shipped the one feature worth borrowing, plus captured rendering lessons from its D3 renderer. Pushed in ff746fe.

## Shipped: Mermaid `erDiagram` export

It's a text output that renders inline in GitHub, Markdown, and chat — which for the MCP agent loop is *better* than either thing `visualize` returned before: a file path needs a shell to open, a URL needs a server the agent can't verify, but a Mermaid block renders where the conversation already is.

- `visualize(nodes:, format: "mermaid")` → fenced `erDiagram` block, directly paste-able.
- `rake woods:map NODES=… FORMAT=mermaid` → writes a `.mmd`.
- Cardinality mirrors rails-schema's proven map, adjusted for our edge orientation (our association edges are FK-holder→PK, so `belongs_to`/`has_many` both render `}o--||` with the macro as label; `}o--o{` for habtm). Non-association code edges render dashed non-identifying (`}o..o{`). Output validated with `mermaid.parse()` end-to-end.

Here's this fixture rendered — GitHub renders it live from the block we emit:

```mermaid
erDiagram
  Order {
    bigint id PK
    bigint account_id FK
  }
  Account {
    bigint id PK
  }
  Invoice {
  }
  OrdersController {
  }
  Order }o--|| Account : "belongs_to"
  Order }o..o{ Invoice : "code_reference"
  OrdersController }o..o{ Order : "code_reference"
```

## Rendering insights from their D3 renderer (for the layout follow-up)

Their `app.js` is a mature force-directed implementation and has concrete answers to the exact problems we hit. Capturing them here so the layout-engine follow-up issue starts from real numbers rather than guesses:

| Problem we had | Their approach (verbatim numbers) |
|---|---|
| **Node overlap** | `forceCollide().radius(max(NODE_WIDTH=180, height)/2 + 50)` — the **+50 buffer** is what actually guarantees separation; charge `-600`, link distance `320`. |
| **Parallel-edge label overlap** (our new `via` pills risk this) | Labels placed at `t=0.25` / `t=0.75` on the Bézier, then a **3-pass** sweep pushing overlapping labels apart by `(overlapY/2)+2`px. |
| **Self-referential edges** | Special-cased loop curves (they don't rely on the graph lib's default) — we currently don't handle self-edges at all. |
| **Degenerate nodes cluttering layout** | Pulled *out* of the simulation and placed deterministically: true orphans in rows above, self-referential-only nodes in a left column. Keeps the force graph clean. |
| **Grouping** | Optional `forceX/forceY` at strength `0.03` toward group centers. |

Two takeaways that change my recommendation slightly:

1. **Their force-directed choice is a caution, not a template, for us.** It's great for a whole-schema overview that clusters by density, but it's a hairball generator for dense graphs and would be *worse* than a hierarchical pass for our *scoped* subgraphs. So the layout follow-up should still be **dagre/ELK (directed)**, not d3-force — but their collide-radius/label-separation math ports directly whichever engine we pick.
2. **The label-separation pass is immediately relevant** — our new `via` pills will collide on parallel association edges at the offsets I added. Worth folding into the layout follow-up.

Also notable: they do **no FK→PK column-level anchoring** (their edges hit the card), which we now do — so on edge precision we're already ahead; on layout we're behind until the follow-up lands.

I'll fold the collide/label numbers and self-edge handling into the layout-engine follow-up issue so it's actionable. Everything else on this PR is unchanged and green (5221 examples, rubocop clean).

---

Issue #164, opened 2026-07-27
Spike: reactive index maintenance — incremental correctness, a resident extraction daemon (`woods:watch`), and multi-instance worktree operation


LeahArmstrong:
# Spike: reactive index maintenance

This is a research + implementation plan for making the Woods index continuously current, rather than as-fresh-as-the-last-explicit-rake-run. It covers three layered change sets: **(0)** correctness fixes the current incremental path needs regardless, **(1)** in-process re-extraction for unit types that today require a full run, **(2)** a resident, file-watching extraction daemon, and **(3)** an MCP freshness contract plus multi-instance operation across parallel git worktrees (worktree managers like [grove](https://github.com/lost-in-the/grove) with agent slots, and multiple concurrent Claude Code sessions).

Nothing here is committed design — each phase names the research questions it must answer before its implementation starts. Phases 0–1 are prerequisites for everything after them.

---

## Problem statement

Woods' freshness model is pull-based: the index changes only when `woods:extract` / `woods:incremental` runs, the MCP Index Server caches on boot (`IndexReader#warmup!`) and refreshes only via the `reload` tool, and staleness signalling is advisory (`woods_status.staleness_seconds` is wall-clock with a 24h threshold; `git_sha_matches_head` only sees committed HEAD, not the working tree). In practice, on any actively-edited checkout, agents consume answers that describe the tree as of the last sync point — and several classes of change never make it into the index at all until a full extraction (see below).

The per-worktree cost model compounds this: each worktree is its own `Rails.root` → its own `tmp/woods` → its own extraction, and every incremental sync from a cold process pays a full Rails boot (typically far more wall-time than the extraction itself). That boot tax is what forces syncing to be batched at hook points (worktree create/switch, session start) instead of continuous.

The differentiating layer of Woods — reflection-true associations, resolved callback chains, inlined concerns, side-effect detection — *requires* a booted app, so "just parse statically" isn't an answer. The thesis of this spike is that the boot requirement doesn't force batch semantics: a **resident booted process** that reloads code and re-extracts affected units on file events can deliver runtime-true answers with seconds of lag. Prior art below.

## Confirmed correctness gaps in the current incremental path (v1.6.1)

[...]

PR #168, opened 2026-07-27
Pre-release fixes: graph path portability, Qdrant point IDs, embed checkpoint durability, GraphQL runtime types

### LINKED ISSUE #147: Qdrant adapter sends Woods identifiers as point IDs — every upsert 400s
## Summary

The Qdrant vector adapter uses Woods unit identifiers directly as Qdrant point IDs, but Qdrant only accepts **unsigned-integer or UUID** point IDs. Every upsert therefore returns HTTP 400 and aborts the embed run, so the `:production` / Qdrant-paired presets cannot store a single vector.

## Evidence (source-verified)

- `lib/woods/storage/qdrant.rb:209` (`store`) and `:240` (`store_batch`) build point bodies with `id: id` / `id: entry[:id]`.
- The IDs come from `lib/woods/embedding/indexer.rb:142` — `embed_id = texts.length > 1 ? "#{identifier}#chunk_#{idx}" : identifier`, i.e. `"User"`, `"Admin::PostsController"`, `"User#chunk_0"`.
- There is no ID-mapping layer anywhere in `lib/woods/storage/`. `search` (`:264`) returns `hit['id']` raw, and `delete` (`:275`) posts the raw identifier.

Existing specs mock HTTP with `'doc1'`-style ids, so they don't catch it; there is no live-Qdrant integration test.

## Suggested fix

- Derive a deterministic **UUIDv5** from the identifier for the Qdrant point ID (stable across runs), and keep the Woods identifier in the point **payload**.
- In `search`, return the payload identifier (not `hit['id']`).
- In `delete`, map the identifier → the same UUIDv5 before posting.
- `delete_by_filter` is unaffected (filter-based).

Add a spec asserting the upsert body carries UUID point IDs and that `search`/`delete` round-trip the Woods identifier. Ideally back it with a live-Qdrant integration lane (currently absent).

## Impact

High — the Qdrant backend is non-functional for real upserts. MySQL stacks (which per `docs/BACKEND_MATRIX.md` **must** pair with Qdrant) have no working vector store today.

Found during the gem audit on `claude/gem-audit-improvements-h64f4x`. Needs adapter-level work + live-backend validation, so filed rather than fixed inline.

### LINKED ISSUE #148: Incremental embed against an in-memory vector store discards new vectors but advances the checkpoint (silent data loss)
## Summary

`rake woods:embed_incremental` against an in-memory vector store (the `:local` / `:shared_filesystem` presets) embeds the changed units into a **throwaway** store, writes nothing to `dumps/`, yet **still advances `checkpoint.json`** — so the next incremental run skips those same units. The changed units' embeddings are unrecoverable without a full `woods:embed`, and the MCP server keeps serving the stale dump in the meantime.

## Evidence (source-verified)

- `lib/woods/embedding/indexer.rb:68-70` — `index_incremental` calls `process_units(load_units, incremental: true)` and returns; it never calls `persist_snapshot` (only `index_all` does, `:62`).
- `process_units` (`:88-101`) unconditionally `save_checkpoint`s after each interval and once at the end, in both modes.
- `tasks.rake`'s embed path rebuilds the vector store fresh each run, so an in-memory store starts empty every time.

Durable backends (pgvector/Qdrant) hydrate/persist externally, which hides the bug in PG-based testing.

## Suggested fix

- In incremental mode, hydrate the in-memory store from the latest dump before processing and `persist_snapshot` after, **or**
- At minimum, do **not** advance the checkpoint when results aren't persisted (so a re-run re-processes the changed units).

Add a spec: incremental embed with a non-durable store must either persist a new dump or leave the checkpoint unchanged for the processed units.

## Impact

High — silent, unrecoverable loss of incremental embeddings on the default local presets.

Found during the gem audit on `claude/gem-audit-improvements-h64f4x`.

### LINKED ISSUE #166: Dependency graph persists absolute file paths, so the index is not portable off the machine that built it
Split out of the #164 / #165 review (finding 17). Deferred from #165 deliberately — it is effectively an index migration and that PR is already at +9,389 lines across 56 files.

## What

`dependency_graph.json` stores **absolute** paths; every per-unit JSON stores **relative** ones. Verified against a real extraction of `spec/dummy`:

```
graph file_map keys (3):
  /tmp/paths20260727-5151-nixeuk/app/models/post.rb
  /tmp/paths20260727-5151-nixeuk/app/models/comment.rb
  /tmp/paths20260727-5151-nixeuk/app/controllers/posts_controller.rb
graph nodes[].file_path (3):
  /tmp/paths20260727-5151-nixeuk/app/models/post.rb
  /tmp/paths20260727-5151-nixeuk/app/models/comment.rb
  /tmp/paths20260727-5151-nixeuk/app/controllers/posts_controller.rb
unit JSON file_path: app/models/comment.rb
```

## Why

Ordering, not intent. In `Extractor#extract_all` the graph is rebuilt from the deduped results at **phase 1.6**, and `normalize_file_paths` (which strips the `Rails.root` prefix) does not run until **phase 4.5**. Registration captures the absolute string; the later reassignment of `unit.file_path` does not reach the copies already in `@nodes` and `@file_map`.

The incremental path has the same ordering in `register_and_write` — register, then normalize — so **full and incremental agree**. This is not an equivalence bug and the #164 harness is right not to flag it.

## Why it matters

1. **The documented Docker deployment.** Extraction runs inside the container and the Index Server runs on the host reading the volume-mounted output (`docs/DOCKER_SETUP.md`). The graph then carries container paths (`/app/...`) while the host sees `/Users/…/woods/...`. Anything that keys on them — `DependencyGraph#affected_by`, `#identifiers_for_path`, the prune sweep — matches nothing from the host side. A host-run `woods:incremental` computes an empty blast radius and silently re-extracts only the dispatched paths.
2. **A moved or renamed checkout** invalidates the whole path index with no error, just a graph that stops resolving.
3. **Two artifacts disagree about the same field.** A consumer reading `file_path` gets an absolute path from the graph and a relative one from the unit, and has to know which artifact it is holding.

## Proposed fix

Normalize **before** registration, so the graph stores relative paths like everything else, and relativize on the way in at the lookup boundary (`ChangeSet` already normalizes to absolute for in-process use — that conversion moves rather than disappears).

## Why it is a migration

Every persisted graph in the wild is absolute-keyed, so `DependencyGraph.from_h` has to detect and convert on load rather than silently mismatch. There is precedent for exactly this — `normalize_edges` handles pre-#164 bare-string edges and `normalize_file_map` handles pre-#164 single-valued file maps, both without a rebuild — so the shape of the fix is established, but it is its own change with its own tests.

Touches: `lib/woods/dependency_graph.rb`, `lib/woods/extractor.rb`, `lib/woods/change_set.rb`, `lib/woods/mcp/index_reader.rb`.

Filed as `B-065` in `docs/backlog.json`.

### LINKED ISSUE #167: GraphQL runtime introspection can emit units no per-file pass reproduces
Follow-up from #164 / #165. The remaining GraphQL divergence after the file-discovery gate was removed in #165.

## What

`GraphQLExtractor#extract_all` has two passes:

- **File pass** — static, a regex over `app/graphql/**/*.rb` plus a `safe_constantize` that is allowed to fail. This is what `PathDispatcher` routes at, via `extract_graphql_file`.
- **Runtime pass** — `extract_from_runtime_type`, driven by the loaded schema when graphql-ruby is present.

#165 made the file pass unconditional, which fixed the divergence that showed up *without* the gem (the per-file path indexed types a full run dropped). What it did not address: with graphql-ruby loaded, the runtime pass can produce units the file pass has no way to reach — types defined dynamically, types built by a schema builder or a DSL rather than a literal `class ... < Types::BaseObject`, and anything whose constant exists only after the schema is constructed.

Those units have no source file, so no changed path ever dispatches to them. An incremental run therefore cannot recreate them, and they persist in the index from whatever full extraction last wrote them — or never appear at all if the index was built incrementally from empty.

## Why it wasn't fixed in #165

It is only observable with graphql-ruby installed, so covering it means adding the gem to the version gemfiles as a dev/test dependency — a six-row change plus a Ruby 3.0 compatibility check for the gem, on a PR already at ~+9,900 lines. The gate fix stands on its own and needs none of that.

## Shape of a fix

Probably the same treatment the other runtime-discovered types get: a `discoverable_classes` equivalent so the incremental path can reconcile the runtime set against the graph in both directions, the way class-based types already do. Worth checking whether the schema's type map is stable enough to serve as that set — if it is not exact by construction, reconciliation would delete units a full extraction still produces, which is the failure mode the class-based gate exists to prevent.

Alternatively GraphQL joins `WHOLE_APP_EXTRACTORS` and re-runs wholesale on any `app/graphql/**` change. Cheaper to reason about, more expensive per change, and it would need the same `gem 'graphql'` test coverage to prove either way.

## Coverage

Needs `gem 'graphql'` in `gemfiles/*.gemfile` and a harness template that exercises a dynamically-defined type. The three templates #165 added cover only the static shapes, which is why they need no gem.

Filed as `B-066` in `docs/backlog.json`.

LeahArmstrong:
## Summary

Four correctness fixes ahead of release. Two are data-loss or total-feature-failure class (`high` severity in the backlog), two close the divergences filed during the #164 review.

| Item | Severity | What was broken |
|---|---|---|
| **#166** / `B-065` | medium | Graph persisted absolute paths — index unusable off the machine that wrote it |
| **#147** / `B-058` | high | Qdrant got Woods identifiers as point IDs — every upsert 400s |
| **#148** / `B-059` | high | Incremental embed discarded vectors but advanced the checkpoint — unrecoverable |
| **#167** / `B-066` | low | Runtime-introspected GraphQL types unreachable from an incremental run |

## Motivation

Closes #166. Closes #167. Fixes #147. Fixes #148.

`B-058` and `B-059` were pre-existing `high` items unrelated to the #164 lineage; they are here because they are genuine release blockers — one makes the Qdrant backend non-functional, the other loses embedded vectors permanently.

### Graph path portability (#166)

The graph held absolute paths in memory **and** on disk, while every per-unit JSON held relative ones. Ordering, not intent: `extract_all` rebuilds the graph at phase 1.6 and `normalize_file_paths` runs at 4.5.

Full and incremental agreed, so the differential harness could never see it — but it broke the documented Docker split. Extraction in the container wrote `/app/...` while the host reading the volume mount saw a different prefix, so a host-run `woods:incremental` computed an empty blast radius and silently re-extracted nothing.

Fixed at the serialization boundary rather than end to end, because mapping the consumers showed that is sufficient: absolute in memory is load-bearing (`re_extract_unit` and `incremental_git_data` gate on `File.exist?`; `ChangeSet` hands `affected_by`/`identifiers_for_path` absolute paths), while **nothing on the read side reads a graph node's `file_path`** — MCP, Obsidian, Unblocked and the formatters all read the per-unit JSON. So `to_h` relativizes, `from_h` absolutizes, and every in-memory consumer is untouched.

Two hazards this turned up:

- `from_h` deliberately takes **no** `root:` keyword. Adding one broke callers passing a bare hash literal — `from_h('nodes' => ...)` binds as keywords the moment the method accepts any, so `data` arrives empty and it raises `ArgumentError`.
- `Watch::Daemon#persisted_registered_paths` is the only consumer of the persisted `file_map`, and its `start_with?(root_prefix)` guard would have excluded every relative entry — silently disabling deletion reconciliation rather than failing visibly.

Stated limitation, spec'd so it isn't mistaken for a bug: a **pre-#166** graph is absolute and cannot be re-pointed at a different root (there is no way to know which prefix was the old machine's). It loads and works in place — which is what avoids a re-index — and becomes portable when a full extraction next rewrites it.

### Qdrant point IDs (#147)

Qdrant accepts only an unsigned integer or a UUID as a point ID; the adapter passed Woods identifiers (`"User"`, `"User#chunk_0"`) straight through, so every upsert 400d and the backend was non-functional.

Point IDs are now UUIDv5 derived from the identifier over a pinned namespace, with the identifier carried in the payload under `woods_identifier`. `#search` reverse-maps so it returns identifiers exactly as pgvector does; `#delete` routes through the same derivation, because a delete computing a different ID is silent data retention rather than a visible error.

Kept inside the adapter — pgvector, the in-memory store and the Indexer are untouched and stay backend-agnostic. The payload key is deliberately not `identifier`: the Indexer already writes that for the unit's *base* id while the point ID derives from the chunk-suffixed embed id.

`POINT_ID_NAMESPACE` must never change — a stable v5 ID is what makes a re-embed replace its point instead of duplicating it. A spec asserts the pinned literal still equals its documented derivation. `Util::UUID5` hashes UTF-8 bytes and transcodes non-UTF-8 names, so an ID does not move with the process's `LANG`.

### Embed checkpoint durability (#148)

On the `:local` / `:shared_filesystem` presets, `index_incremental` never called `persist_snapshot` while `process_units` advanced `checkpoint.json` regardless. Changed units embedded into a store the process then threw away, nothing reached `dumps/`, and the checkpoint skipped them on the next run — unrecoverable without a full re-embed.

Reproduced before fixing, across three runs against one index:

```
full run:        {processed: 1}              dump ids: ["User"]   ckpt: {User}
incremental #1:  {processed: 1, skipped: 1}  dump ids: ["User"]   ckpt: {User, PaymentService}
incremental #2:  {processed: 0, skipped: 2}  dump ids: ["User"]
```

`PaymentService` was embedded, paid for, checkpointed — and permanently absent.

The invariant now enforced: **the checkpoint never advances over a unit whose vector was not durably stored.** Incremental runs hydrate from `dumps/latest` and dump before the checkpoint is written; `checkpoint_interval` saves are suppressed on that path (a dump is a whole-store snapshot, so there is no partial durability for them to record) while durable backends keep them, since there each `store_batch` *is* the durable write. `checkpoint_satisfied?` honours a hash match only when the artifact actually holds that vector, so a checkpoint that ran ahead self-heals into a re-embed rather than stranding the unit.

### GraphQL runtime types (#167)

A type existing only at runtime — built by a schema builder or DSL rather than a literal `class ... < Types::BaseObject` — has no source file, so no changed path could dispatch to it. Only a full extraction emitted those units.

`GraphQLExtractor#discoverable_classes` now exposes the runtime set and GraphQL joins `CLASS_BASED_DISCOVERY`. The load-bearing part is `reconcile_removals: false`, which GraphQL alone sets: its unit type is the **union** of two independent discovery mechanisms, so "in the graph but not in `discoverable_classes`" does not mean deleted. Without graphql-ruby loaded the runtime set is empty and removal would wipe every GraphQL unit; with it loaded, a file-defined type not attached to the schema is legitimately absent from the type map. Additions are safe in both worlds and are exactly the divergence.

**This shipped inert in `033699e` and was fixed in `d4d14d8`.** `extract_from_runtime_type` was defined below `private`, and `add_discovered_classes` invokes it with `public_send` — so every candidate raised `NoMethodError`, the reconciler's `rescue StandardError` turned it into a warn and a nil, `filter_map` dropped it, and the incremental path added nothing. Neither pre-existing guard could see it: the table spec compares the symbol, which stays true while the method is private, and the reconcile specs inject a double, which answers `public_send` regardless of the real class's visibility. A new spec checks visibility statically for every `CLASS_BASED_DISCOVERY` entry — no gem, no booted app, and it covers future entries.

Documented consequence of the opt-out: a runtime-only type that *disappears* is never removed by an incremental run and survives until a full extraction rebuilds the type. That is the right side to err on — a stale unit is recoverable by one full run, whereas the removal half would delete units a full extraction still emits.

graphql-ruby remains absent from the gemspec and the version gemfiles.

## Testing

- `bundle exec rake spec` — **5,458 examples, 0 failures.** Pending count is environment-dependent (2 here, 4 on the Ruby 3.0 row and on the reviewer's host) and is entirely tiktoken benchmarks skipping when `tiktoken_ruby` is absent — no real skips.
- `bundle exec rubocop` — clean, 520 files
- Differential harness (`:booted_app`) — 17/17 at `WOODS_DIFF_OPS=60` × seeds 1–3, run after each change that touches the incremental path
- UUIDv5 verified against two published vectors (`UUIDv5(DNS, "python.org")`, `UUIDv5(DNS, "www.example.com")`) rather than against itself, plus encoding-invariance
- #148 verified by reproducing the loss before fixing, and by temporarily forcing interval checkpoints back on to confirm the guard fails
- The #167 visibility guard confirmed to fail on `033699e` and pass on `d4d14d8`

### Live validation — complete

All three gaps originally flagged as unvalidated have since been closed against a real stack (details in the review thread):

1. **Live Qdrant 1.18.3, 18/18.** A real server confirms the premise verbatim — `value User is not a valid point ID, valid values are either an unsigned integer or a UUID`. Upserts accepted; two further re-upserts of unchanged units leave the count flat rather than doubling; `#search` returns Woods identifiers with no UUID leaking; `#delete` drops exactly one point; namespace still equals its derivation.
2. **#148 upgrade-repair**, with corruption generated by running `origin/main` rather than hand-edited. Re-embeds only the lost unit, warns once, and the follow-up run is a silent no-op — so the heal is durable.
3. **#166 at app scale** in the Docker split it describes: 5,961 absolute paths before, 5,821 relative after, with the 140 remaining absolute all `rails_source`/`gem_source` living outside the app root.

## Checklist

- [x] Tests added/updated
- [x] `bundle exec rake spec` passes
- [x] `bundle exec rubocop` passes
- [x] CHANGELOG.md updated
- [x] Documentation updated — CLAUDE.md gotchas for all four; `docs/BACKEND_MATRIX.md` for Qdrant; `B-058`, `B-059`, `B-065`, `B-066` marked resolved

https://claude.ai/code/session_019rcyqTCoi5FE5QzVaNXNPR


LeahArmstrong:
Ran the three items listed as "not validated here" against a live stack, plus the round 1 surface. #147, #148 and #166 hold up. **#167 shipped inert**, so I pushed a fix to this branch as `d4d14d8`.

## #167 was a no-op

`CLASS_BASED_DISCOVERY` names `:extract_from_runtime_type` as the graphql entry's method and `add_discovered_classes` invokes it with `public_send`, but the method was defined below `private`. Every candidate raised `NoMethodError`. That is a `StandardError`, so the reconciler's rescue turned it into a warn and a nil, `filter_map` dropped it, and the incremental path added nothing. That is exactly the case #167 set out to fix.

Measured against a real schema (graphql-ruby 2.5.9) with a runtime-only type attached to the query root and no file under `app/graphql`:

| gem | graphql units | type in graph | warnings |
|---|---|---|---|
| `033699e` as shipped | 312 to 312 | no | one per candidate, every run |
| visibility patched | 312 to 313 | yes | none |
| `d4d14d8` | 313 to 314 | yes | none |

Neither existing guard could catch it. The table spec compares the symbol, which stays true while the method is private. The reconcile specs inject a double, and a double answers `public_send` for whatever is stubbed on it regardless of the real class's visibility. The comment on that block already calls this out as "a guard on the wiring, not a behavioural test of a live reconcile."

The new spec checks visibility statically for every `CLASS_BASED_DISCOVERY` entry, so it needs neither graphql-ruby nor a booted app, and it covers future entries. Confirmed it fails on `033699e` and passes on `d4d14d8`. Suite 5458/0, rubocop clean.

<details><summary><b>Evidence for #147, #148, #166 and the round 1 surface</b></summary>

**#147, live Qdrant 1.18.3, 18/18.** Checked the premise first, and a real server confirms it verbatim: `value User is not a valid point ID, valid values are either an unsigned integer or a UUID`. Then: batch and single upserts accepted; two further re-upserts of unchanged units leave the count flat at 4 rather than doubling; `#search` returns Woods identifiers with no UUID leaking through; the chunk payload keeps `identifier` and `woods_identifier` distinct; filtered search works; `#delete` drops exactly one point without taking the base unit; `delete_by_filter` works; `POINT_ID_NAMESPACE` still equals its documented derivation.

**#148, upgrade-repair, with corruption produced by the shipping code.** `573f858^` is `c1d63cb`, which is current `origin/main`, so the damage was generated by running main rather than by hand-editing `checkpoint.json`. It reproduced the signature in the PR body exactly:

```
full run:        {processed: 1}              dump=["User"]   ckpt=[User]
incremental #1:  {processed: 1, skipped: 1}  dump=["User"]   ckpt=[User, PaymentService]
incremental #2:  {processed: 0, skipped: 2}  dump=["User"]
```

Pointing that same index dir at this branch re-embeds only the lost unit (`processed: 1, skipped: 1`, not a full re-embed), warns exactly once, and leaves `dump=["PaymentService", "User"]`. The follow-up run is a silent no-op, so the heal is durable.

**#166, at app scale, in the Docker split it describes.** The pre-fix index had 5961 of 5961 node paths absolute under the container prefix. After a full extraction on this branch: 5821 relative, and zero app files still absolute. The 140 that remain absolute are all `rails_source` (116) and `gem_source` (24), living outside the app root, and they are byte-identical to the previous index because `extract_framework` owns them. The incremental path writes relative paths too: the unit added in the #167 test landed with a relative `file_path`.

**Round 1 surface.** The watch daemon reaches running, polls at 1s, reconciles startup deletions, and completes an incremental to a new generation. The freshness contract holds: `Incremental run changed nothing, leaving manifest timestamp untouched`.

Two things I checked and cleared rather than reporting. `woods:watch_status` exiting 1 while the daemon is alive is correct, because `alive?` guards on `same_host?` and I was querying from a different container. In-container `woods:incremental` silently no-ops inside a git worktree, because the worktree's `.git` points at a gitdir the container cannot see, but that is already documented and worked around host-side.

</details>

## Smaller notes, no action needed

**#167 does not manifest in this app today.** All 193 named runtime types are file-defined; the only 9 without a source file are graphql-ruby builtins, which `extract_from_runtime_type` skips on the `GraphQL::` prefix. The fix is correctness insurance here rather than a live bug. One side effect: those 9 can never enter the graph, so they are re-attempted on every incremental run. Post-fix they are silent nils, but pre-fix that was 10 warn lines per run.

**`reconcile_removals: false` has a visible consequence.** A runtime-only type that disappears is never removed by an incremental run. I watched the probe units survive deletion of the file that defined them, and only a full extraction cleared them. That follows from the documented reasoning, and is worth stating explicitly next to the opt-out so it does not read as a bug later.

**Pending count.** The PR body says 2 pending; the suite reports 4, all tiktoken.

LeahArmstrong:
Thanks — and the #167 catch is squarely mine. I shipped a change whose entire purpose was to make the incremental path add those units, and it added nothing. Worse, it failed in the quietest possible way: `NoMethodError` is a `StandardError`, so the reconciler's rescue turned every candidate into a warn and a nil that `filter_map` dropped. Exactly the "silently contributes nothing" failure mode this PR fixes elsewhere.

Your fix and its guard are better than what I wrote, so I've taken `d4d14d8` as-is. I verified the guard independently rather than trusting the claim: restored `033699e`'s `graphql_extractor.rb`, and it fails with `graphql -> Woods::Extractors::GraphQLExtractor#extract_from_runtime_type`; restored the fix, and it passes.

The part worth keeping is *why* it went undetected. I wrote the table spec myself and described it in a comment as "a guard on the wiring, not a behavioural test of a live reconcile" — so I knew it was weak and still let it stand as the only static coverage. Checking the symbol matches while the method is unreachable is a test of my own table against itself. Your static visibility check is the version that has teeth, and it costs neither the gem nor a booted app.

## Acted on both notes

Pushed as `41bce27`.

**`reconcile_removals: false` cost.** Now stated next to the opt-out, in the code and in CLAUDE.md: a runtime-only type that disappears is never removed by an incremental run and survives until a full extraction rebuilds the type. Your probe-unit observation is recorded as the evidence. I also noted why it is the right side to err on — a stale unit is recoverable by one full run, whereas the removal half would delete units a full extraction still emits.

**The public-method rule** is recorded too, with why neither pre-existing guard caught it, so the new spec is not later mistaken for redundant with the symbol-matching one.

**Pending count.** You are right that the body was wrong, though the fix is not a number: it is environment-dependent. This container reports 2, your run and the CI Ruby 3.0 row both report 4 — all tiktoken benchmarks, none a real skip. I have changed the body to say so rather than swap one machine-specific number for another.

## On the two you cleared

Both worth having in writing, since they will come up again: `woods:watch_status` exiting 1 from a different container is `same_host?` doing its job, and the worktree `.git` gitdir issue is the documented one. Good calls not to file them.

Everything else is now measured rather than asserted — live Qdrant on the actual `not a valid point ID` error, #148 repaired from corruption generated by `origin/main` rather than hand-edited, and #166 at 5,821 relative paths with the 140 remaining absolute correctly accounted for as framework sources outside the app root. That closes all three gaps I flagged as unvalidated.

Suite 5,458, 0 failures. RuboCop clean.

---

Issue #181, opened 2026-07-29
pgvector batch upsert raises PG::CardinalityViolation when a batch contains a duplicate id


LeahArmstrong:
## Description

Embedding an extracted index into pgvector fails partway through:

```
Woods::Error: Embedding failed: PG::CardinalityViolation: ERROR:
  ON CONFLICT DO UPDATE command cannot affect row a second time
HINT: Ensure that no rows proposed for insertion within the same command
  have duplicate constrained values.
```

PostgreSQL raises this when a single `INSERT … ON CONFLICT DO UPDATE` statement proposes **two rows with the same conflicting key**. So `Storage::VectorStore::Pgvector`'s batch upsert is being handed a batch containing the same id twice, and it does not deduplicate before building the statement.

Earlier batches land successfully — a `search` after the failure does return hits — so the failure is data-dependent and leaves the index **partially embedded**, which is the worse half: there is no error until some batch happens to contain a collision, and what's already stored stays.

## Steps to Reproduce

Against `woods-testbed`'s `rails-8.0-large` variant (1,940 units), with the `backends` compose profile up:

```ruby
store = Woods::Storage::VectorStore::Pgvector.new(
  connection: VectorDb.connection,     # an ActiveRecord connection
  dimensions: 128
)
store.ensure_schema!

Woods::Embedding::Indexer.new(
  provider: deterministic_provider,
  text_preparer: Woods::Embedding::TextPreparer.new,
  vector_store: store,
  output_dir: Rails.root.join("tmp/woods")
).index_all
# => PG::CardinalityViolation
```

The full script is `scripts/woods_embedding_smoke.rb` in that repo.

## Expected Behavior

`index_all` completes, or fails with a message naming the duplicated id.

## Actual Behavior

`PG::CardinalityViolation`, a partially populated `woods_embeddings` table, and no indication of which id collided.

## Where the duplicate probably comes from

I have not isolated the source, so treating this as two candidate causes rather than asserting one:

1. **Chunk id collision** — a unit large enough to be chunked contributes several rows; if the per-chunk id is not unique (or a unit is chunked twice within one batch) the same key appears twice.
2. **The known identifier collision (B-062)** — the graph keys nodes on the bare identifier, so two units of different types sharing one collapse onto a single node. The same collision would put two rows with one id into an embedding batch. If that is the cause, this is B-062 surfacing in the storage layer, which would make it a second, louder symptom of an already-known problem.

[...]

Issue #150, opened 2026-07-01
Audit follow-ups: medium-severity correctness/robustness findings (server, notion, unblocked, resilience, bootstrap)


LeahArmstrong:
Consolidated medium-severity findings from the gem audit (branch `claude/gem-audit-improvements-h64f4x`). Each is source-verified but lower-impact or needs more design than the high-severity issues (#147–#149). Splitting into separate PRs is fine.

> **Update (PR #151):** items **2, 3, 4** below are now **fixed** on the audit branch with regression tests. Items **1** and **5** remain open (they touch concurrency/backend-load paths better validated with a live index/transport).

## 1. HTTP transport runs tool handlers concurrently, but shared state is unsynchronized — ⚠️ PARTIAL
`lib/woods/mcp/server.rb` — the mcp gem's HTTP dispatch runs outside the stdio server's mutex. The lazy `@pipeline_mutex ||= Mutex.new` init race is **fixed** (mutex now eagerly initialized). The remaining piece is `IndexReader`'s cache check-then-read, which is still unsynchronized → possible spurious "unit not found" / broken LRU eviction under concurrent reads. Fix: guard the IndexReader cache.

## 2. Notion / Unblocked clients retry non-idempotent POSTs on `Net::ReadTimeout` — ⏳ still open
`lib/woods/notion/client.rb`, `lib/woods/unblocked/client.rb` — a POST that times out after the server committed is retried, creating duplicate pages/collections the upsert never reconciles. Fix: don't retry POST/PATCH on read-timeout, or make creation idempotent via an external key. *(Note: distinct from the Retry-After fix already shipped.)*

## 3. `Retry-After` parsed with `.to_f` — ✅ FIXED (PR #151)
Fixed via a shared `Woods::RetryAfter` helper that parses both delta-seconds and HTTP-date forms; both clients use it.

## 4. `CircuitBreaker` half-open admits unlimited concurrent probes — ✅ FIXED (PR #151)
Half-open now admits a single probe at a time (concurrent calls rejected with `CircuitOpenError`), which also removes the overlap reset race. Added `success_threshold` (default 1) for consecutive successes to close.

## 5. Shape-2 boot backfill / MCP `reload` write string-keyed (or absent) vector metadata while `SearchExecutor` filters with symbol keys — ⏳ still open
[...]

Issue #11, opened 2026-03-25
Feature request: Factory/fabricator extraction


LeahArmstrong:
## Summary

Test factories (FactoryBot, Fabrication) define canonical data shapes and valid object graphs but aren't currently extracted. In our project we have 22 Fabrication fabricators that document how to construct valid test data — including traits, associations, sequences, and transient attributes.

## Use case

Factories are the authoritative answer to "what does a valid X look like?" — more so than the model itself, since factories encode realistic attribute combinations and required associations. Having them in the index would help AI assistants:

- Reference existing fabricators when generating tests instead of constructing data from scratch
- Understand which traits and transient attributes are available
- Map fabricator associations to model relationships
- Avoid creating redundant test setup that duplicates existing fabricators

## Prior art

The [use cases and feature gaps doc](https://github.com/lost-in-the/woods/blob/main/docs/USE_CASES_AND_FEATURE_GAPS.md) documents this as Gap 7 (low-medium impact), noting that AI test generation cannot reference existing factories without manual source reads.

## Suggested scope

- Extract fabricator/factory definitions with their model association
- Surface traits, transient attributes, sequences, and association declarations
- Link fabricators to their corresponding model units
- Support both FactoryBot and Fabrication gems

LeahArmstrong:
Closing as implemented. `FactoryExtractor` (`lib/woods/extractors/factory_extractor.rb`) has covered this for a while: a full parser for factories, traits, associations, sequences, callbacks, and parent inheritance, registered in `Extractor::EXTRACTORS`, refreshed wholesale via `WHOLE_APP_EXTRACTORS`, spec-covered, and documented in `docs/EXTRACTOR_REFERENCE.md`. The 2026-07-30 full-gem review audited the request against the implementation and found it satisfied.

One residue is tracked separately: blockless factories (`factory :admin, parent: :user`) are currently dropped — see #215.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #10, opened 2026-03-25
Feature request: Stimulus controller extraction


LeahArmstrong:
## Summary

Stimulus controllers are a core part of Hotwire-based Rails apps but aren't currently extracted or indexed. In our project we have 43 Stimulus controllers under `app/javascript/controllers/` that define significant UI behavior — live reporting, moderation interfaces, drag-and-drop, form handling, real-time connection monitoring — with no representation in the Woods index.

## Use case

When working with AI assistants, understanding the JavaScript behavioral layer is essential for:
- Tracing user interactions end-to-end (controller action → Turbo Stream → Stimulus controller)
- Understanding which `data-controller` attributes map to which behaviors
- Identifying dependencies between Stimulus controllers and ActionCable channels
- Cross-referencing view templates with their behavioral counterparts

## Prior art

The [coverage gap analysis doc](https://github.com/lost-in-the/woods/blob/main/docs/COVERAGE_GAP_ANALYSIS.md) mentions this as medium-effort work well-suited to JS AST parsing, noting maximum value when cross-referenced with view template extraction (which Woods already does well — 181 templates indexed in our project).

## Suggested scope

- Extract controller name, targets, values, actions, outlets
- Link to view templates that reference `data-controller="<name>"`
- Surface connect/disconnect lifecycle hooks

PR #210, opened 2026-07-30
Full-gem review: fix 30 verified defects across extraction, retrieval, storage, coordination, and exports

### LINKED ISSUE #149: Notion Columns/Data-Models sync collapses same-named columns across models (data corruption)
## Summary

The Notion Columns sync upserts each column page keyed by the bare column name and finds existing pages by **title equality across the whole Columns database**. Because every Rails model shares `id` / `created_at` / `updated_at`, syncing a second model overwrites the first model's page — the Columns DB churns down to one page per *distinct column name in the entire app*, with the last-synced model's Data Type / Nullable / Validation / `Table` relation winning.

## Evidence (source-verified)

- `lib/woods/notion/exporter.rb:146` upserts with `title_value: column['name']` (bare, e.g. `"id"`).
- `upsert_page` (`:180-190`) locates existing pages by title equality only, with no `Table`-relation qualifier.
- `lib/woods/notion/mappers/column_mapper.rb:29` confirms the bare-name title.
- Same collision hits Data Models for STI subclasses keyed by shared `Table Name`.

Specs stub `find_page_by_title` to always return nil, so the collision path never runs.

## Suggested fix

- Qualify the column page title (e.g. `users.id`), **or**
- Filter the existing-page lookup by the `Table` relation in addition to the title.

Add a spec that syncs two models sharing a column name and asserts two distinct pages result.

## Impact

High for anyone using the Notion export with more than one model — the synced database is silently corrupted.

Found during the gem audit on `claude/gem-audit-improvements-h64f4x`.

### LINKED ISSUE #169: rails_source should be a Gemfile.lock-keyed whole-app extractor; woods:extract_framework bypasses the write pipeline entirely
**Rewritten after review.** The original framing was wrong: all four options I listed assumed the fix was calling `woods:extract_framework` more often. That is the wrong lever, because the task bypasses the write pipeline — so running it more often reproduces the bypass on a schedule.

## The actual defect

`rails_source` and `gem_source` units are written **only** by an explicit `rake woods:extract_framework`. Nothing automatic runs it — not the sync script, not the watch daemon, not `woods:incremental`, and not `woods:extract`. Once written they persist unchanged forever.

But the deeper problem is *how* that task writes. `lib/tasks/woods.rake:321-346` dumps `JSON.pretty_generate(unit.to_h)` straight into `rails_source/<name>.json` and skips every step a real extraction takes:

| Step | `extract_framework` |
|---|---|
| `normalize_file_paths` (`extractor.rb:901`) | skipped |
| `regenerate_type_index` → `_index.json` | skipped |
| manifest counts + `gemfile_lock_sha` | skipped |
| `publish_generation` | skipped |
| `woods_with_extraction_lock` | **never taken** |

That first row is the direct cause of the observed symptom: the 140 units carry absolute paths because they never went through `normalize_file_paths`, not merely because they are old. The last row makes this a **fifth unlocked writer** against the index — the `refresh` task's own comment (`woods.rake:300`) counts four and misses this one.

### Evidence

On a real host, 140 units (116 `rails_source`, 24 `gem_source`) carry host-style paths belonging to a *different worktree*, while the surrounding index was extracted in a container where the equivalent path is `/app/vendor/bundle/...`. Those paths resolve nowhere, host or container. They were also the only units left absolute after #166 — the same symptom from the other side.

## The fix

`Gemfile.lock` is **already** a dispatch trigger. `path_dispatcher.rb:178-180`:

```ruby
whole_app_rule(:engines,    %w[config/routes], exact_paths: %w[config/routes.rb Gemfile.lock]),
whole_app_rule(:middleware, %w[config/initializers config/environments],
               exact_paths: %w[config/application.rb Gemfile.lock]),
```

So "refreshes wholesale, only when the lockfile moves" already exists and is in use. Make `rails_source` a member:

- add `rails_source: :rails_source` to `WHOLE_APP_EXTRACTORS` (`extractor.rb:249`)
- add `whole_app_rule(:rails_source, [], exact_paths: %w[Gemfile.lock])`
- delete `woods:extract_framework`, or reduce it to a thin `refresh[rails_source]` alias

`WHOLE_APP_EXTRACTORS` is not "extractors that are expensive" — it is "extractors whose output is a function of the whole app, so an incremental run replaces their output wholesale." That is exactly the contract framework sources need. The comment at `extractor.rb:245` excludes `rails_source` because it "changes only on a dependency bump", but the mechanism is keyed on *dispatch rules*, so an extractor appearing only in a `Gemfile.lock` rule gets that behaviour for free. The exclusion bought nothing the rule would not have given.

### Why this beats the options originally listed

- **vs "the sync script runs it"**: no cost on the hook path for runs that don't touch `Gemfile.lock`, and no second trigger surface competing with the gem's dispatcher.
- **vs "digest check in `woods:extract`"**: a whole-app extractor runs on every full extraction by construction, so no digest comparison is needed. It also self-corrects only on full extraction and leaves the bypass in place.
- **vs "prune on mismatch"**: wholesale replacement already drops what no longer exists.
- It makes the sync script's existing "run a full extraction" message *true* rather than requiring a rewrite of what it promises, and the 140 stale units clear on the next `woods:extract` with no migration.

**Keep from the original option 3:** the `woods:validate` unresolvable-`file_path` check. Cheap, and it catches this class generally rather than this instance.

## Decision needed by whoever implements it

`normalize_file_path` (`extractor.rb:914`) only strips the `Rails.root` prefix and returns everything else untouched — the docstring explicitly calls out gem paths as "not under Rails.root". Where the bundle is vendored at `vendor/bundle` **under** `Rails.root`, framework paths relativize and become portable across worktrees. With a global `GEM_HOME` they stay absolute and remain non-portable. So "relativize framework units" is not a free consequence of routing them through the pipeline; it needs a decision about what root gem paths are relative to, or explicit acceptance that they are valid only in the extracting environment.

## Related, not in this repo

The `woods-sync` script lives in the consuming application's repository, not here. Two fixes belong with this work, on that side:

1. Its full-extraction-only path filter mirrors `Woods::Sync::FULL_EXTRACT_ONLY_PATTERNS`, a constant #164/#165 **deleted** — `lib/woods/sync.rb` no longer exists. The semantics diverged with it: under the current dispatcher `config/routes.rb` refreshes `:routes` and `:engines`, and `Gemfile.lock` refreshes `:engines` and `:middleware`, but the script strips both out of the incremental call. So a `Gemfile.lock` change currently *loses the engines and middleware refresh that would have worked*, on top of the framework units that would not have. Worth checking `db/structure.sql` the same way.
2. Its "clear the sync cursor and re-run a full extraction" message needs correcting either way.

## Also unlocked, same class

- `woods:clean` (`woods.rake:504-512`) is a bare `FileUtils.rm_rf(output_dir)` — run mid-cycle it removes the lock file itself, so the holder's `release` hits `ENOENT` and the next contender acquires immediately. Deliberate human action, so low frequency.
- `woods:embed` / `woods:embed_incremental` are unlocked, though arguably a separate concurrency domain since the daemon does not embed.

If this issue is fixed as recommended, the unlocked-`extract_framework` problem dissolves — the task stops existing. Fixed any other way, it needs handling separately.

Filed as `B-067` in `docs/backlog.json`.

### LINKED ISSUE #170: woods:clean and the embed tasks write against the index unlocked
Split out of #169, where it was recorded but does not belong — the mechanism is unrelated to framework staleness, so it would have been closed alongside a fix that does not address it.

## `woods:clean`

`lib/tasks/woods.rake:504-512` is a bare `FileUtils.rm_rf(output_dir)` with no `woods_with_extraction_lock`.

Run while the daemon is mid-cycle it removes **the lock file itself**, so the holder's `release` hits `ENOENT` (handled, no error) and the next contender acquires immediately. The result is two writers plus a daemon publishing into a deleted tree.

Deliberate human action, so the frequency is low — but the failure is silent and the recovery is a full re-extraction.

The fix is not simply "take the lock", since the thing being deleted contains the lock. Options: take the lock, delete everything *except* the lock file, then release and remove it; or refuse to run while a daemon is alive (`Status#alive?`) the way `woods:incremental` already stands down, with `WOODS_IGNORE_WATCH=1` as the override.

## `woods:embed` / `woods:embed_incremental`

Also unlocked. Arguably a separate concurrency domain — the daemon does not embed, so these do not contend with it today. Worth an explicit decision rather than leaving it implicit: if embedding ever moves under the daemon, or if two embed runs can overlap on one index, the same clobber applies. `PipelineGuard`'s cooldown reduces the window for full runs but does not serialise them, and incremental runs are not rate-limited at all.

## Context

The audit that produced the lock heartbeat asked "who refreshes the lock" and found three gaps. The prior question — "who takes the lock at all" — has its own answers: `woods:extract_framework` (covered in #169, and it dissolves if that is fixed as recommended), plus these.

Filed as `B-068` in `docs/backlog.json`.

### LINKED ISSUE #174: Namespace lost in class-name detection: StateMachineExtractor and ServiceExtractor index namespaced classes under their inner name
## Description

Two extractors derive a unit identifier from the **innermost** class name in a file, discarding the enclosing module. A namespaced class is therefore indexed under a name that does not match its constant, and two same-named classes in different namespaces collapse onto one graph node.

Affected call sites:

- `StateMachineExtractor#detect_class_name` — `lib/woods/extractors/state_machine_extractor.rb`
- `ServiceExtractor` (its `extract_class_name` path) — `lib/woods/extractors/service_extractor.rb`

The class-based extractors are unaffected, because they take `model.name` from the runtime constant. This is specific to the extractors that parse a name out of source.

## Steps to Reproduce

Against `woods-testbed`'s `rails-8.0-large` variant (which carries both shapes deliberately):

1. `app/models/billing/payment.rb` containing `module Billing; class Payment < ApplicationRecord` with an `aasm column: :state do … end` block.
2. `app/use_cases/billing/issue_invoice.rb` containing `module Billing; class IssueInvoice`.
3. `bin/rails woods:extract`

```ruby
Woods::Extractors::StateMachineExtractor.new
  .send(:detect_class_name, File.read("app/models/billing/payment.rb"), "app/models/billing/payment.rb")
# => "Payment"
```

```ruby
JSON.parse(File.read("tmp/woods/services/_index.json")).map { _1["identifier"] }
# => ["IssueInvoice", "PublishArticle"]
```

## Expected Behavior

- State machine unit identifier: `Billing::Payment::aasm`
- Service unit identifier: `Billing::IssueInvoice`

Matching `ModelExtractor`, which correctly emits `Billing::Invoice` and `Billing::LineItem` for the same tree.

## Actual Behavior

- State machine unit identifier: `Payment::aasm`
- Service unit identifier: `IssueInvoice`

Two consequences beyond the wrong name:

1. **Cross-namespace collision.** `Billing::Payment` and (say) `Legacy::Payment` both yield `Payment::aasm`. `DependencyGraph` keys nodes on the bare identifier, so the second registration wins — the same collapse recorded as B-062, reached by a second route. `deduplicate_results` won't catch it because both units are the same type.
2. **Dependency edges can't resolve.** Anything referencing the correct constant has no node to point at, so the unit is effectively orphaned in the graph.

## Notes

Found while building a kernel fixture that deliberately puts a state machine on a *namespaced* STI base. A flat, single-namespace fixture cannot surface this, which is why it has survived — worth a regression spec with a namespaced class for each of the two extractors.

## Environment

- Ruby version: 3.3.1
- Rails version: 8.0.5
- Database adapter: sqlite3
- Woods version: `main` (`c1d63cb`)
- OS: Linux (Docker, `ruby:3.3.1-slim`)

### LINKED ISSUE #175: Controller concerns are never recorded: extract_included_concerns matches on the module name containing "Concern"
## Description

`ControllerExtractor#extract_included_concerns` identifies a controller's concerns by checking whether the **module's name contains the string** `"Concern"` or `"Concerns"`:

```ruby
# lib/woods/extractors/controller_extractor.rb
def extract_included_concerns(controller)
  controller.included_modules
            .select { |m| m.name&.include?('Concern') || m.name&.include?('Concerns') }
            .map(&:name)
end
```

Rails does not namespace controller concerns — `app/controllers/concerns/requires_author.rb` defines a top-level `RequiresAuthor`. So an idiomatically named controller concern is invisible, and `included_concerns` comes back `[]` for a controller that includes one.

Only a module whose name happens to contain the substring (`FooConcern`, or something nested under a literal `Concerns::`) is ever detected.

## Steps to Reproduce

1. `app/controllers/concerns/requires_author.rb`:

```ruby
module RequiresAuthor
  extend ActiveSupport::Concern
  included { before_action :require_author }

  private

  def require_author
    head :forbidden if Author.find_by(id: session[:author_id]).nil?
  end
end
```

2. `app/controllers/articles_controller.rb` with `include RequiresAuthor`.
3. `bin/rails woods:extract`
4. Read `tmp/woods/controllers/ArticlesController_*.json`.

## Expected Behavior

```json
"included_concerns": ["RequiresAuthor"]
```

## Actual Behavior

```json
"included_concerns": []
```

The concern's *effect* is picked up — `filters` correctly contains `{"kind": "before", "filter": "require_author"}` — so the extraction knows the behaviour arrived, just not where from. `ancestors` is also `["ArticlesController", "ApplicationController"]`, with the module absent.

## Suggested fix

Test membership rather than the name. Either:

- the module is defined under one of the `app/**/concerns` load paths (`Object.const_source_location`), or
- `m.singleton_class.include?(ActiveSupport::Concern)` — i.e. the module was extended by `ActiveSupport::Concern`

The second is cheap and catches the idiomatic case directly. `ConcernExtractor` already discovers the concern *files* correctly via `Dir[Rails.root.join('app/**/concerns')]`, so the two could agree on one definition.

## Related: controller concern source is not inlined

Separate but adjacent — model units get `metadata.inlined_concerns` **and** the concern's source spliced into `source_code`, verified on the same tree:

| Unit | metadata | source contains concern body |
|---|---|---|
| `Article` (model) | `inlined_concerns: ["Archivable"]` | yes — `module Archivable`, `def archive!` |
| `ArticlesController` | `included_concerns: []` | no |

That asymmetry may well be deliberate; the gem's CLAUDE.md only ever claims inlining "when extracting a model". Flagging it because "concerns get inlined" is stated as a design decision without qualification, and a reader of controller output would reasonably expect the same. Happy to split this half into its own issue if you'd rather keep them separate.

## Environment

- Ruby version: 3.3.1
- Rails version: 8.0.5
- Database adapter: sqlite3
- Woods version: `main` (`c1d63cb`)
- OS: Linux (Docker, `ruby:3.3.1-slim`)

### LINKED ISSUE #176: RakeTaskExtractor misses the modern `task name: :environment` form, so most real .rake files yield no units
## Description

`RakeTaskExtractor#parse_task_signature` recognises three task shapes, and all three require a **leading colon** on the task name:

```ruby
# lib/woods/extractors/rake_task_extractor.rb
line.match(/\Atask\s+:(\w+)\s*,\s*\[([^\]]*)\]/)          # task :name, [:args]
line.match(/\Atask\s+:(\w+)\s*=>\s*(.+?)(?:\s+do|\s*$)/)  # task :name => :deps
line.match(/\Atask\s+:(\w+)/)                             # task :name
```

The Ruby 1.9 hash form — `task archive_stale: :environment` — has no leading colon and matches none of them. `parse_task_signature` returns `nil` and the task is dropped.

This is the form most Rails codebases actually use, so the practical effect is that `rake_task` extraction silently returns few or no units on a typical host app.

## Steps to Reproduce

`lib/tasks/publishing.rake`:

```ruby
namespace :publishing do
  desc "Report article counts per author"
  task :report => :environment do          # old form
    Author.find_each { |a| puts a.name }
  end

  desc "Archive articles unpublished for a year"
  task archive_stale: :environment do      # modern form
    Article.where(published_at: nil).find_each(&:archive!)
  end
end
```

```ruby
Woods::Extractors::RakeTaskExtractor.new.extract_all.map(&:identifier)
# => ["publishing:report"]     # archive_stale missing
```

With only the modern form present, the result is `[]` — and because the directory *is* found and the file *is* read, there is no warning to indicate anything was skipped.

## Expected Behavior

Both tasks extracted: `publishing:report` and `publishing:archive_stale`.

## Actual Behavior

Only the hash-rocket task is extracted. The modern-form task is absent from the index with no diagnostic.

## Suggested fix

Add a branch for the label-syntax form, e.g. `\Atask\s+(\w+):\s*(.+?)(?:\s+do|\s*$)/` — being careful to order it after the `task :name, [:args]` branch so an args list isn't misread. The dependency parsing on the right-hand side is identical to the existing hash-rocket branch, so `parse_dependency_list` can be reused unchanged.

Worth a spec covering all four forms together, since the failure mode here is silent.

## Environment

- Ruby version: 3.3.1
- Rails version: 8.0.5
- Database adapter: sqlite3
- Woods version: `main` (`c1d63cb`)
- OS: Linux (Docker, `ruby:3.3.1-slim`)

### LINKED ISSUE #177: Full extraction leaves orphaned unit files behind, so unit counts over-report and woods:validate warns
## Description

`Extractor#setup_output_directory` creates directories but never clears them:

```ruby
# lib/woods/extractor.rb
def setup_output_directory
  FileUtils.mkdir_p(@output_dir)
  EXTRACTORS.each_key do |type|
    FileUtils.mkdir_p(@output_dir.join(type.to_s))
  end
end
```

So a full extraction into an output directory that has seen a *different* version of the app writes the current units over the top and leaves the previous run's unit files in place. `manifest.json` and each `_index.json` are rewritten correctly, but the stale `<Identifier>_<hash>.json` files remain.

Incremental extraction prunes vanished units deliberately (#164). A full extraction has no equivalent step, which is the opposite of what you'd expect from the two modes.

## Steps to Reproduce

1. Extract an app containing `Account` and `AccountEvent`.
2. Delete those models, add different ones, and extract again (full, no `woods:clean`).
3. `bin/rails woods:validate`

## Expected Behavior

The index describes only the current tree; `woods:validate` is clean.

## Actual Behavior

```
WARNINGS:
  ⚠ models: expected 10, found 12
  ⚠ controllers: expected 3, found 4
  ⚠ routes: expected 16, found 18
  ⚠ view_templates: expected 10, found 12
  ⚠ migrations: expected 2, found 4

✓ Index is valid with 5 warning(s).
```

`expected` is the manifest count (correct); `found` is the file count on disk (inflated by orphans). `woods:clean && woods:extract` produces a clean validate on the identical tree, which confirms the cause.

## Why it matters beyond the warning

Anything that reads the index by globbing a type directory rather than through `_index.json` sees units for classes that no longer exist. Concretely:

- **Benchmark and stats numbers over-report.** We hit this building a scale variant for [woods-testbed#2](https://github.com/lost-in-the/woods-testbed/issues/2) — the whole point of that variant is trustworthy unit counts and whole-app re-run percentages, and a dirty output directory silently inflates both.
- **`woods:validate` warns rather than fails**, so a CI job that only checks the exit code won't notice.

## Suggested fix

Either prune each type directory of files not written by the current run (the run already knows the full written set — `write_results` could collect written paths and sweep the difference), or clear the type directories in `setup_output_directory`. The first is safer: it keeps `dumps/`, `woods.json`, and other non-unit artifacts alone, which a blanket clear would not.

Until then, documenting "run `woods:clean` before a full extraction into a reused directory" would be enough to stop the over-reporting silently.

## Environment

- Ruby version: 3.3.1
- Rails version: 8.0.5
- Database adapter: sqlite3
- Woods version: `main` (`c1d63cb`)
- OS: Linux (Docker, `ruby:3.3.1-slim`)

### LINKED ISSUE #178: Embedding pipeline can't be driven from a host app without a live provider, and woods:retrieve hardcodes Ollama + InMemory
## Description

Two related gaps that together make the embedding half of the gem untestable in a host app, and put one rake task in conflict with the backend-agnostic rule in CLAUDE.md.

### 1. `Builder` accepts only the two live providers

```ruby
# lib/woods/builder.rb
def build_embedding_provider
  case @config.embedding_provider
  when :openai then Embedding::Provider::OpenAI.new(**opts)
  when :ollama then Embedding::Provider::Ollama.new(**opts)
  else raise ArgumentError, "Unknown embedding_provider: #{@config.embedding_provider}"
  end
end
```

`woods:embed` and `woods:embed_incremental` both go through `Woods::Tasks.build_embed_indexer` → `Builder`, so **there is no way to run `rake woods:embed` in a host app without a reachable OpenAI or Ollama endpoint.** That rules it out of CI and out of any sandboxed or offline environment.

A deterministic provider already exists and is exactly what's needed — `Woods::Embedding::Provider::Fake` in `spec/support/fake_embedding_provider.rb`, bag-of-words hashing with L2 normalisation so cosine similarity stays meaningful. It just isn't reachable from outside the gem's own suite.

`Embedding::Indexer` already takes `provider:` as a kwarg, so a smoke script *can* construct the pipeline directly with any provider object. That's the workaround we're using in `woods-testbed`, but it means the actual `woods:embed` → `woods:retrieve` chain stays untested end to end.

### 2. `woods:retrieve` hardcodes its backends

```ruby
# lib/tasks/woods.rake
provider       = Woods::Embedding::Provider::Ollama.new
vector_store   = Woods::Storage::VectorStore::InMemory.new
metadata_store = Woods::Storage::MetadataStore::SQLite.new
graph_store    = Woods::Storage::GraphStore::Memory.new
```

It reads `Woods.configuration` for `max_context_tokens` on the next line but ignores the configured `embedding_provider`, `vector_store`, `metadata_store` and `graph_store` entirely. So on a host configured for OpenAI + pgvector, `woods:retrieve` queries a different backend than `woods:embed` wrote to — and silently returns nothing useful rather than erroring.

This contradicts CLAUDE.md's own rule: *"Backend agnostic. … Never hardcode or default to a single backend."*

## Expected Behavior

1. A provider can be supplied without a network endpoint — either `embedding_provider = :fake` recognised by `Builder`, or `Builder` accepting an object that responds to the `Provider::Interface` contract (which would also let hosts plug in their own).
2. `woods:retrieve` resolves all four backends through `Builder` / `Woods.configuration`, like `build_embed_indexer` does.

## Actual Behavior

1. `rake woods:embed` raises or hangs without a live endpoint.
2. `rake woods:retrieve` always queries Ollama + in-memory regardless of configuration.

## Notes

Surfaced while planning storage/embedding coverage for [woods-testbed#2](https://github.com/lost-in-the/woods-testbed/issues/2). We've verified that pgvector and Qdrant both run and are reachable in a containerised session, so the only thing standing between the testbed and real round-trip coverage of `embed` → `retrieve` is the provider injection point. Promoting `Provider::Fake` out of `spec/support/` into `lib/` would be the smallest change that unblocks it.

Happy to send a PR for either half if the shape above sounds right — particularly whether you'd prefer a `:fake` symbol or object injection.

## Environment

- Ruby version: 3.3.1
- Rails version: 8.0.5
- Database adapter: sqlite3 (host), pgvector/Qdrant available as containers
- Woods version: `main` (`c1d63cb`)
- OS: Linux (Docker, `ruby:3.3.1-slim`)

### LINKED ISSUE #179: CLAUDE.md describes the old lock-wait behaviour: says manual rake runs proceed with a warning, code exits non-zero
## Description

CLAUDE.md's `PipelineLock` gotcha still describes the behaviour that `woods_with_extraction_lock` deliberately replaced.

**CLAUDE.md** (Gotchas, the `PipelineLock` bullet):

> Manual rake runs wait 30s then proceed with a warning (hanging CI is worse than an overlap the index survives).

**`lib/tasks/woods.rake`**, in the comment on `woods_with_extraction_lock`:

> This used to proceed *without* the lock after 30s, on the reasoning that a daemon cycle is milliseconds so a longer wait meant something unusual. That reasoning was wrong in the case that matters: a cycle includes a storm-triggered `extract_all`, which on a large host app runs for minutes. […] So the wait is now generous and the failure explicit. `WOODS_LOCK_WAIT` overrides it; exceeding it exits non-zero rather than corrupting the index.

And the code:

```ruby
wait ||= Float(ENV.fetch('WOODS_LOCK_WAIT', Woods::Watch::Daemon::LOCK_STALE_TIMEOUT))
```

So the current behaviour is: wait for `LOCK_STALE_TIMEOUT` (overridable via `WOODS_LOCK_WAIT`), then **fail loudly** — not 30 seconds then proceed.

This matters more than a normal doc nit because the two statements recommend opposite things, and the parenthetical in CLAUDE.md ("an overlap the index survives") asserts exactly what the rake comment says was wrong — an unlocked concurrent write *doesn't* survive, since the file set isn't atomic. An agent reading only CLAUDE.md could reintroduce the behaviour on purpose.

## Suggested fix

Replace the sentence with something like:

> Manual rake runs wait `LOCK_STALE_TIMEOUT` for the lock (override with `WOODS_LOCK_WAIT`) and then **exit non-zero** rather than proceeding unlocked — a storm-triggered `extract_all` can hold the lock for minutes on a large host, and two concurrent writers silently discard one another's graph.

Same bullet also still reads "all four of them", which remains correct.

## Environment

- Woods version: `main` (`c1d63cb`)

### LINKED ISSUE #183: Railtie console-MCP wiring 401s the entire host app, and both middleware enable flags are read before config/initializers run
## Description

Two railtie defects that together make the documented console-MCP / session-tracer setup either a silent no-op or an outage:

1. **Unscoped guards.** `lib/woods/railtie.rb:68` inserts `Woods::MCP::OriginGuard` and `Woods::MCP::BearerAuth` via `app.middleware.use` with no path scoping. `BearerAuth#call` (bearer_auth.rb:33-42) 401s every request lacking the bearer header; `OriginGuard#host_allowed?` 403s any non-loopback `Host`. Only `Console::RackMiddleware` is path-aware, and it sits *below* the guards. With the feature enabled, `GET /` returns 401; under a real hostname all traffic gets 403 even with the token. Verified against a booted Rails 8.1 app.

2. **Early flag read.** `railtie.rb:13` / `:60` evaluate `console_mcp_enabled` / `session_tracer_enabled` at railtie-initializer time, which runs before `:load_config_initializers` (observed tsort positions 50 vs 63). Setting either flag in `config/initializers/woods.rb` — the file the install generator creates and every doc names — can never mount the middleware. No warning, no error.

## Suggested fix

Path-scope the guards (a `path:` kwarg passing through non-matching paths, or compose them inside `Console::RackMiddleware` after its own path check), and decide enablement lazily at request time via the existing `enabled?` check — or warn from `after_initialize` when a flag is set but nothing was mounted.

Tracked as **B-071** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #184: QueryClassifier target-type patterns match ordinary English words and hard-filter vector search with no fallback
## Description

`TARGET_PATTERNS` (`lib/woods/retrieval/query_classifier.rb:56`) treat very common words as type signals — `\bget\b` → `:route`, `\bemail\b` → `:mailer`, `\bquery\b`/`\btype\b`/`\bfield\b` → `:graphql`, `\bperform\b` → `:job` — and `SearchExecutor#build_vector_filters` (search_executor.rb:302) pushes the classifier-derived `target_type` down as a **hard** metadata filter on the vector search. Unlike caller-supplied `types:`, there is no within-type fallback (`Retriever#apply_type_filter` falls back only for caller `type_list`, retriever.rb:318).

Reproduced mechanically: `How do we get the current user?` → `:route` filter, so the auth concern, ApplicationController, and User are all excluded and the primary section can come back empty. `How is the email field stored?` → `:mailer`. The Ranker already implements the soft version of this signal (`type_match`, weight 0.10), so the hard filter is redundant when right and destructive when wrong.

## Suggested fix

Stop pushing classifier-derived `target_type` into the vector filter (keep the pushdown for explicit caller `types:`); let the Ranker's `type_match` express the heuristic. If a hard filter is kept, reuse the existing `within_type_fallback` machinery on empty results and tighten the worst patterns.

Tracked as **B-072** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #185: Ranking layer is largely inert: dead matched_fields signal, unnormalized RRF, assembler re-sort discards ranked order, PageRank memo never invalidated on reload
## Description

Four compounding defects mean the documented weighted-signal ranking rarely determines what an agent actually receives:

1. **`matched_fields` is dead code** (`ranker.rb:174`). `keyword_score` guards on `candidate.matched_fields`, but `SearchExecutor::Candidate` is `Struct.new(:identifier, :score, :source, :metadata)` — no such member exists anywhere in `lib/` (only in a design doc). Every candidate scores 0.0; `WEIGHTS[:keyword] = 0.20` is permanently inert.
2. **Raw RRF into 0–1 weights** (`ranker.rb:97`). `apply_rrf` scores max ≈ 0.016/source, so the 0.40 semantic weight contributes ≤ ~0.02 while the recency spread alone is 0.105 — fused relevance is swamped by metadata signals on every hybrid query. Reproduced with the real Ranker.
3. **Assembler re-sort** (`context_assembler.rb:222`). The Ranker's output is ordering-only; `assemble_section` re-sorts by raw `candidate.score`, so recency/importance/type_match/diversity never influence what enters the context or what budget truncation cuts.
4. **Stale PageRank memo** (`ranker.rb:229`). `@pagerank_importance_map` is memoized forever; `Bootstrapper.reload_stores!` swaps the graph in place and invalidates the context cache but never touches the memo — a long-lived `woods-mcp-http` serves boot-time percentiles after a successful `reload`.

## Suggested fix

Populate `matched_fields` from the keyword path (which knows the probed fields) or delete the signal and renormalize; min-max normalize RRF across the merged list before weighting; have the Ranker rewrite `candidate.score` to the weighted score (as `apply_rrf` already does) or make the assembler preserve input order; add `Ranker#invalidate_pagerank_cache!` called from `reload_stores!`.

Specs: a ranker-demoted candidate stays demoted in the assembled context; the fused-relevance winner survives divergent recency/type metadata; reload picks up new PageRank. Relatedly, the evaluation harness that would have surfaced all of this has never been runnable from a shipped entry point — worth fixing alongside (#178 covers part).

Tracked as **B-073** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #186: ContextAssembler assembles the same framework unit into both :primary and :framework sections
## Description

Section membership is not disjoint (`lib/woods/retrieval/context_assembler.rb:112-118`): the primary section takes `candidates.reject { |c| c.source == :graph_expansion }` — which includes vector-sourced `rails_source`/`gem_source` candidates — and the framework section takes `candidates.select { |c| framework_candidate?(c) }` over the same list. A framework unit retrieved by vector search satisfies both, so its full source is formatted twice in one context, consuming both budgets for identical content. `sources.uniq` (line 359) hides the duplication from the attribution list while the context string still carries it twice.

This fires precisely in the mainline case the framework section exists for: a framework-context query ("how does activerecord save work") whose top hits are `rails_source` units — up to a third of an 8k-token budget on one duplicated blob, evicting app-code candidates. Reproduced end-to-end through the real pipeline.

## Suggested fix

Partition instead of overlapping selects: exclude `framework_candidate?` units from the primary section whenever the framework budget is active. Spec: each identifier's source text appears exactly once in the assembled context.

Tracked as **B-074** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #187: pgvector backend is unusable via the documented path: ensure_schema! has no callers and the woods:pgvector generator creates dead schema
## Description

The Pgvector adapter reads/writes its own `woods_vectors` table, created only by `Pgvector#ensure_schema!` (`lib/woods/storage/pgvector.rb:35`) — which nothing in the gem ever calls. `Builder#build_vector_store` (`builder.rb:132`) instantiates the adapter bare. Meanwhile the documented setup step, `rails generate woods:pgvector && rails db:migrate`, creates an `embedding_vector` column + HNSW index on `woods_embeddings` — a column with zero readers or writers anywhere in `lib/`.

So a user who configures `vector_store = :pgvector` (or `configure_with_preset(:postgresql)`) and follows every documented step gets `PG::UndefinedTable: relation "woods_vectors" does not exist` on the first embed, and the migration they ran provides no protection.

## Suggested fix

Call `store.ensure_schema!` from `Builder#build_vector_store` (or lazily on first write) for the pgvector branch, and either delete the `woods:pgvector` generator or rewrite its template to create `woods_vectors` with the dimension option so the schema is migration-managed for the table actually used.

Related: #181 (in-batch duplicate-id dedup for the same adapter).

Tracked as **B-075** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #188: Resilience layer is never wired — a single 429 aborts woods:embed and discards paid-for work
## Description

`RetryableProvider` and `CircuitBreaker` have zero production callers (grep: only `lib/woods/resilience/` and docs). `Tasks.build_embed_indexer` (`lib/woods/tasks.rb:42`) and `Builder#build_retriever` hand the **raw** provider to the pipeline, so any OpenAI 429/5xx (or an Ollama hiccup beyond the single connection-drop retry) raises `Woods::Error` and aborts the whole run.

On the `:local`/`:shared_filesystem` presets interval checkpoints are suppressed by design and the dump only lands at end-of-run — so a mid-run rate limit discards **every embedding already paid for in that run**. First full embed of a 10k-unit host that hits an org TPM limit at batch 200/400 restarts from zero every attempt, spending API budget each time.

This directly violates the layer's own rule file (`.claude/rules/storage-retrieval.md`: "Never let a rate limit crash the indexing pipeline"), and `docs/USE_CASES_AND_FEATURE_GAPS.md` claims the breaker protects embedding — it protects nothing.

## Suggested fix

Wrap the provider in `Resilience::RetryableProvider.new(provider:, circuit_breaker: CircuitBreaker.new)` in both build sites, and honor `Retry-After` on 429 (the shared `Woods::RetryAfter` helper already exists). Acceptance: a provider raising 429 twice then succeeding completes the run.

Tracked as **B-076** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #189: Bare File.read crashes woods:embed, Unblocked sync, and StatusReporter under LANG=C — AtomicFile.read convention breaches
## Description

CLAUDE.md mandates `AtomicFile.read` for Woods' own JSON artifacts precisely because `File.read` tags results with the process default external encoding (US-ASCII under `LANG=C`, the plain-Docker default). Three satellites still violate it, and two are reproduced crashers:

1. **`Indexer#load_units`** (`lib/woods/embedding/indexer.rb:91`) — bare `File.read` + `rescue JSON::ParserError` only. Any non-ASCII byte in a unit's `source_code` (an em dash in a comment) raises `Encoding::InvalidByteSequenceError` and crashes the whole `woods:embed` run before embedding anything. Reproduced under forced US-ASCII external encoding. `load_checkpoint` (:489) shares the breach and `save_checkpoint` (:495) is also non-atomic.
2. **`Unblocked::SyncManifest#load`** (`lib/woods/unblocked/sync_manifest.rb:114`) — same pattern; one non-ASCII URI in the manifest and every subsequent sync crashes at `Exporter` construction until the file is manually deleted, violating the class's own "corrupt file degrades to everything-is-new" contract. Reproduced.
3. **`Operator::StatusReporter`** (`lib/woods/operator/status_reporter.rb:51`) — reads `manifest.json` with bare `File.read`, rescuing only `JSON::ParserError`.

## Suggested fix

`AtomicFile.read`/`.write` at all three sites; widen rescues to include `EncodingError` on the degrade paths; add non-ASCII fixtures loaded under forced US-ASCII external encoding (the suite's US-ASCII canary doesn't reach these paths today). Consider a small custom cop banning bare `File.read`/`File.write` under `lib/woods/` so the convention stops regressing.

Tracked as **B-077** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #190: Flow layer bypasses AtomicFile and persists absolute container paths into unit JSON and flow_index.json
## Description

Two convention breaches in `FlowPrecomputer` that undermine the #166 portability contract and the documented degradation path:

1. **Absolute paths persisted** (`lib/woods/flow_precomputer.rb:32,50-57,98-101`): `@flows_dir` is built from the absolute output dir, and those absolute paths are written into `flow_index.json` values and `unit.metadata[:flow_paths]`, which `Extractor#rewrite_flow_annotated_units` persists into per-unit JSON. Extraction inside Docker bakes `/app/tmp/woods/flows/...` into artifacts a host-side reader serves verbatim; two extractions of identical trees from different roots produce different bytes — defeating the `canonical_json` churn-avoidance the class itself implements.
2. **Non-atomic writes, narrow read rescue** (`flow_precomputer.rb:87,100`; read side `lib/woods/mcp/server.rb`): flow docs and `flow_index.json` are the only index artifacts written with plain `File.write`, and the MCP server reads them with bare `File.read` rescuing only `JSON::ParserError`/`Errno::ENOENT`. An `Encoding::InvalidByteSequenceError` under `LANG=C` (UTF-8 literals in `args_hint`/`condition` strings sliced from app source) escapes to the generic handler, so `trace_flow` returns `internal_error` instead of degrading to query-time reassembly as documented.

## Suggested fix

Store output_dir-relative paths in `flow_map` and `metadata[:flow_paths]` (readers join against their own index_dir — `load_precomputed_flow` already does); write via `AtomicFile.write`, read via `AtomicFile.read`, and widen the read rescue to `EncodingError` so any unreadable precomputed flow degrades to reassembly.

Tracked as **B-078** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #191: drop_vanished_units has no mass-deletion guard: an empty load_units wipes the entire vector index and rotates away good dumps
## Description

If an incremental embed run sees zero unit JSONs — mismatched `WOODS_OUTPUT` between the extract and embed shells, extraction output deleted while `dumps/` survives, a permissions/glob failure — `@current_identifiers` stays empty, so **every hydrated identifier counts as vanished**: `drop_vanished_units` (`lib/woods/embedding/indexer.rb:193`) prunes all vectors and `persist_snapshot` promotes an effectively empty dump as `latest`. With the default retention of 3, two more accidental runs evict every dump that still held real data — the full embedding spend is unrecoverable.

Reproduced with the real Indexer + snapshotter: 5 units across 3 dumps, unit JSONs deleted, one `index_incremental` run → all vectors pruned, empty dump promoted.

The gem's other destructive sweeps (Obsidian, Unblocked) are explicitly gated behind an ownership sentinel and a 30% purge guard; this one has only a post-hoc warn.

## Suggested fix

Refuse (or require an explicit override) when `@current_identifiers` is empty but `@persisted_ids` is not, and add a proportional guard — abort the prune when vanished > 30% of persisted unless overridden — warning **before** deleting rather than after.

Tracked as **B-079** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #192: WVF1 idx ids hydrate as ASCII-8BIT: non-ASCII identifiers never match their UTF-8 twins — perpetual re-embed and duplicate live entries
## Description

`parse_idx` (`lib/woods/storage/snapshotter/vector.rb:128`) builds ids with `String#byteslice` on a binread buffer, so hydrated ids are ASCII-8BIT-tagged. For ASCII ids that's invisible, but a non-ASCII identifier (e.g. a path-derived view_template id with an accented segment) is not `eql?` to the same bytes tagged UTF-8, and hash lookups miss. The consequences chain through the Indexer:

- `@persisted_ids.key?(utf8_id)` is false → the checkpoint self-heal re-embeds the unit on **every incremental run forever** (with a warn each time);
- `InMemory#store` appends a second live entry instead of overwriting (id_to_index lookup misses), so the store and each subsequent dump accumulate one duplicate per run and search returns the unit twice.

The existing spec explicitly asserts *byte* equality rather than encoding equality (`vector_spec.rb:454`), papering over exactly this.

## Suggested fix

`id.force_encoding(Encoding::UTF_8)` after the byteslice in `parse_idx` — the idx format already documents UTF-8 bytes (vector.rb:28) — and tighten the spec to `eq(original_id)` including encoding.

Tracked as **B-080** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #193: Model concern inlining silently skipped for compact-style class definitions; CallbackAnalyzer can never see concern-defined callbacks
## Description

Two defects that hollow out the gem's stated key differentiator ("concerns get inlined") for common code shapes:

1. **Compact-style skip** (`lib/woods/extractors/model_extractor.rb:221`): `build_model_source_with_concerns` inserts the inlined-concern block via `source.sub(/(class\s+<demodulized name>.*$)/)`. A model defined compact-style — `class Library::Book < ApplicationRecord` — never matches `class\s+Book` (verified in plain Ruby), so `sub` no-ops and the computed concern block is **silently discarded**, while `metadata.inlined_concerns` (:340) is computed independently and still claims the concerns were inlined. The regex also lacks a trailing `\b`, so for model `Book` a file defining `class BookmarkError` first gets the block inserted after the wrong class. Compact style is fully Zeitwerk-supported and rubocop-endorsed (`Style/ClassAndModuleChildren: EnforcedStyle: compact`).

2. **Comment-blind analyzer** (`model_extractor.rb:210`, `:677`): concern source is inlined as `# `-prefixed comment lines, and that composite is handed to `CallbackAnalyzer`, whose Prism parse cannot see commented code. The standard idiom — `included do; before_save :set_slug; end` with `def set_slug` in the concern — always yields empty `side_effects`, contradicting the analyzer's own docs ("works on composite source code (with inlined concerns)").

## Suggested fix

Match both forms — `/(class\s+(?:[\w:]+::)?Name\b.*$)/` — with an append-plus-warn fallback when insertion fails, and derive `inlined_concerns` from whether insertion actually happened. Give `CallbackAnalyzer` a parse-friendly composite (raw model + raw concern sources); keep the commented form for display only.

Tracked as **B-081** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #194: TestMappingExtractor mis-parses `RSpec.describe Klass, type: :model` — the rspec-rails generator default
## Description

`extract_rspec_subject`'s constant regex (`lib/woods/extractors/test_mapping_extractor.rb:126`) — `/^\s*(?:RSpec\.)?describe\s+([\w:]+)\s/` — requires whitespace immediately after the constant, so the generator-standard form `RSpec.describe User, type: :model do` (comma after the constant) never matches. The string fallback (:130) then matches the **first quoted describe anywhere in the file** — typically an inner `describe 'validations' do` — and that string becomes `subject_class` and a `:model` dependency target via `infer_type_from_test_type`.

Reproduced in plain Ruby: the constant regex returns nil (the `\s` cannot match the comma and backtracking can't rescue it), the fallback returns `"validations"`, and `extract_dependencies` emits `{type: :model, target: "validations", via: :test_coverage}` — the real `User` coverage edge is lost and a ghost graph node is minted. Since explicit type metadata is the rspec-rails default, most real host apps get a broken test→code mapping plus graph noise. No spec covers the `, type:` form.

## Suggested fix

Allow a trailing comma or `do` — `/^\s*(?:RSpec\.)?describe\s+([A-Z][\w:]*)[\s,]/` — and anchor the string fallback to the file's first describe line (line-by-line scan) rather than `.match` over the whole source. Add specs for the `, type: :model` form and string-only files.

Tracked as **B-082** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #195: Chunkers close methods at the first nested if/case/begin end; endless defs never close
## Description

All three chunkers (`ControllerChunker#track_action_line`, `MethodChunker#track_method_line`, `ModelChunker#update_method_depth` — `lib/woods/chunking/semantic_chunker.rb:461` and siblings) increment depth only for `do` blocks (ModelChunker also `def`), never for multi-line `if`/`unless`/`case`/`begin`/`while`. The `end` closing such a nested block decrements depth to 0 and terminates the current method chunk — so every line after the nested block, plus the method's real closing `end`, is misclassified into the `:summary` chunk with wrong `chunk_type` metadata.

Endless methods (`def show = render json: @post`, Ruby 3.0+) are worse: method-tracking state opens with no `end` ever coming, so all subsequent methods are swallowed into that one chunk.

Reproduced with a standalone script running the real chunker: `def create; if @post.save; redirect_to @post; end; NotifyJob.perform_later(@post); end` — the job enqueue lands in `:summary`, and the embedded `action_create` chunk omits it, so retrieval for "job enqueued on create" matches the wrong chunk.

## Suggested fix

Track depth with the same opener set `RakeTaskExtractor#block_opener?` already uses (line-leading `if`/`unless`/`case`/`begin`/`while`/`until` plus `do`/`def`) — or better, replace the line-regex state machines with the existing Prism-based `Ast::MethodExtractor` and slice chunks by AST line ranges. Don't enter method-tracking state when the def line is an endless def. Specs: the action chunk contains the method's final `end` and code following a nested `if`.

Tracked as **B-083** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #196: Standing-down daemon clobbers the live daemon's status file with 'stopped' and empties watch_pending.json
## Description

`Daemon#run`'s ensure block runs on the early `return :already_running` path (`lib/woods/watch/daemon.rb:191`), so a second daemon that **correctly stands down** for a live one still executes `persist_pending` (overwriting `watch_pending.json` with `[]`) and `publish_status(:stopped, ...)` — overwriting the live daemon's `running` record with `state: stopped` under the stander-down's pid.

Reproduced end-to-end: after a stand-down, the status file reads `stopped`/dead-pid and `watch_pending.json` is emptied. For up to `HEARTBEAT_INTERVAL` (300s, until the live daemon restamps):

- `woods:watch_status` exits 1, so the documented hook shape `rake woods:watch_status || start_daemon` immediately starts a **third** daemon — whose `another_daemon_alive?` reads the `stopped` record and proceeds, producing exactly the two-live-daemons state the guard exists to prevent;
- `woods:incremental` stops standing down and contends with the live daemon's lock;
- carried-forward paths recorded in `watch_pending.json` are lost if the live daemon is later SIGKILLed.

## Suggested fix

Return before any side-effecting state, guarded so the ensure doesn't fire — e.g. set `@started = true` after the alive-check and wrap the ensure body in `if @started`, or restructure as `return :already_running if another_daemon_alive?; run_started(...)` with the ensure inside the inner method. Spec: the live daemon's status record and `watch_pending.json` are byte-identical after a stand-down.

Tracked as **B-084** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #197: PipelineLock#touch can resurrect a just-released lock as an ownerless phantom file
## Description

`touch` uses `FileUtils.touch(@lock_path)` (`lib/woods/coordination/pipeline_lock.rb:163`), which **creates** the file when it does not exist. The daemon's heartbeat thread calls `@lock.touch` concurrently with the drain thread's `@lock.release` at the end of an extraction cycle. If release completes (rename + rm_f) between the heartbeat's `lock_ownership` read (which returned `:ours`) and its `FileUtils.touch`, the touch recreates the lock path as a fresh empty file that **no process owns and no `release` will ever delete**.

Every writer — the daemon itself, `woods:extract`, `woods:incremental` — then sees a fresh, non-stale lock and blocks until the full 600s stale window expires; manual rake runs wait `WOODS_LOCK_WAIT` then exit 1 claiming another writer holds the lock while none exists. The realistic trigger is a storm-triggered full extraction running ~5 minutes, so the 300s heartbeat tick fires mid-run and races the cycle-end release. Reproduced mechanically.

## Suggested fix

Refresh without create semantics: replace `FileUtils.touch(@lock_path)` with `t = Time.now; File.utime(t, t, @lock_path)` — `ENOENT` is already covered by the method's `rescue SystemCallError => false`. Optionally re-verify ownership by opening the file and fstat-ing the same fd to close the window entirely.

Tracked as **B-085** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #198: Incremental safety gates applied in one place but not their twin: failed extractor construction returns [] not nil; replace_type_wholesale ignores @eager_load_complete
## Description

Two cases where a documented safety principle is enforced on one path and missing on its sibling, both ending in silent mass deletion — exactly the failure mode the principles exist to prevent:

1. **`[]` vs `nil` on failed construction** (`lib/woods/extractor.rb:1524`): `extract_with_rule` carefully returns `nil` (not `[]`) when per-file extraction raises, so a transient failure cannot license pruning. But when the extractor **instance** fails to construct, `extractor_for` (:1449) rescues and memoizes `nil`; `extract_with_rule` then hits `return [] unless extractor.respond_to?(rule.method_name)` — nil doesn't respond — and returns `[]`, the "this path defines nothing anymore" answer. A watcher batch naming 12 view templates while `ViewTemplateExtractor#initialize` transiently raises prunes all 12 units, reports success, and bumps the generation.

2. **Ungated wholesale removal** (`extractor.rb:1722`): `stale_class_based_units` opens with `return [] unless @eager_load_complete` — documented as "the whole safety argument" — but `replace_type_wholesale` performs the same discovery-set-driven removal with no gate, and the `ROUTE_CONSUMER_EXTRACTORS` cascade routes four descendants-discovered types through it on every routes change. On a NameError-fallback boot (discovery sets known-partial), one routes edit mass-deletes most controller units.

## Suggested fix

(1) `return nil if extractor.nil?` in `extract_with_rule` — construction failure tells us nothing and must not license pruning; keep `[]` only for a constructed extractor lacking the method. Spec: stub `EXTRACTORS[key].new` to raise, assert existing units survive. (2) Gate the removal loop in `replace_type_wholesale` on `@eager_load_complete` for keys in `CLASS_BASED_DISCOVERY` (file-derived whole-app types keep unconditional removal); warn when removal is skipped.

Tracked as **B-086** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #199: Polymorphic associations emit phantom :model dependency edges
## Description

`extract_dependencies` (`lib/woods/extractors/model_extractor.rb:631`) maps every association reflection to `{type: :model, target: assoc.class_name, via: assoc.macro}` without checking `assoc.polymorphic?`. For `belongs_to :commentable, polymorphic: true`, Rails derives `class_name` `"Commentable"` — an interface name, not a model (`AssociationReflection#class_name` camelizes without constantizing, so the existing `rescue NameError` never fires).

Consequences: with a `Commentable` **concern** in `app/models/concerns` (polymorphic interfaces are conventionally named after concerns), the graph wires `Comment → Commentable-concern` as a `:model`/`:belongs_to` edge and dependents-of-Commentable conflates includers with polymorphic owners; without such a concern the edge dangles at a node no extraction produces.

## Suggested fix

When `assoc.polymorphic?`, emit a distinct edge — e.g. `via: :polymorphic_interface` carrying the interface name — or resolve concrete types from other models' `as:` options, instead of a `:model` edge on the derived class_name.

Tracked as **B-087** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #200: ApplicationController — and direct ActionController::Base subclasses — are never indexed
## Description

`discoverable_classes` (`lib/woods/extractors/controller_extractor.rb:49-53`) is `ApplicationController.descendants (+ ActionController::API.descendants)`. `Class#descendants` excludes the receiver, so **ApplicationController itself** — typically the richest controller in the app (authentication filters, `rescue_from` handlers, concern includes, layout logic) — is never extracted in a full-stack app. Controllers inheriting directly from `ActionController::Base` (health endpoints, engine base controllers) are also invisible.

The unguarded constant reference is a second bug: a host without `ApplicationController` raises `NameError` out of `discoverable_classes`, which `reconcile_class_based_types` calls without rescue (`extractor.rb:1596`).

Observable effects: retrieval for "what does ApplicationController do / where is authenticate_user! defined" returns nothing; child controllers' `ancestors` metadata names an identifier that resolves to no unit.

## Suggested fix

Discover from `ActionController::Base.descendants + ActionController::API.descendants` (both `defined?`-guarded), filtering framework-internal classes by app-source path — so app-defined base controllers and direct Base subclasses are included and API-only hosts don't raise.

Tracked as **B-088** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #201: CachingExtractor assigns the file's first cache key/TTL to every cache_calls entry
## Description

`extract_cache_calls` (`lib/woods/extractors/caching_extractor.rb:166-178`) iterates pattern occurrences, but for each occurrence calls `extract_key_pattern(source, type)` and `extract_ttl(source)` against the **whole source** — both use `String#match`, which returns the first match in the file. Every `cache_calls` entry therefore reports the same key_pattern and ttl: those of the first `Rails.cache` call and first `expires_in` in the file.

Reproduced with the extractor's verbatim regexes: a controller with `Rails.cache.fetch("products/index")` and `Rails.cache.write("cart/#{'{'}id{'}'}", ..., expires_in: 5.minutes)` produces two entries, **both** with key `"products/index"` and **both** with ttl `5.minutes` — wrong key and TTL attribution for any file with more than one cache call, which is the extractor's mainline case.

## Suggested fix

Iterate with MatchData positions (`source.to_enum(:scan, pattern)` + `Regexp.last_match`, or manual `match(pattern, offset)`) and parse the key and any `expires_in:` from that call's own argument text rather than the whole file. Spec: two calls produce two distinct key_patterns/TTLs.

Tracked as **B-089** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #202: GraphQLExtractor#extract_class_name appends inner classes to the identifier, defeating runtime/file dedup
## Description

`extract_class_name` (`lib/woods/extractors/graphql_extractor.rb:366-373`) joins **every** `module`/`class` declaration in the file (`modules.join('::')`), so a nested helper or error class inside a mutation/type is appended to the identifier. Reproduced: `module Mutations / class CreateUser < BaseMutation / class InvalidInput < StandardError` yields `Mutations::CreateUser::InvalidInput`.

Because `extract_all` dedups the file pass against the runtime pass **by identifier** (:95-:107), the wrongly-named file unit no longer collides with the runtime unit's correct name:

- with graphql-ruby loaded → the same class is indexed twice under two identifiers (inflated counts, duplicate search results);
- without the gem → only the wrong identifier exists, so dependency edges targeting `Mutations::CreateUser` dangle.

Inline error classes are a common graphql-ruby pattern, so this hits ordinary schemas. Part of the same family as #174 (namespace derivation from source, wrong in a third distinct way).

## Suggested fix

Stop at the first `class` declaration: join only the module declarations that precede it plus that class name — the same fix shape as the PORO/Lib half of the #174 family, which argues for one shared position-aware nesting parser. Spec: a mutation with an inner error class asserts the outer identifier.

Tracked as **B-090** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #203: I18n and scheduled-job YAML with anchors/aliases is silently dropped (Psych 4 AliasesNotEnabled)
## Description

`I18nExtractor` (`lib/woods/extractors/i18n_extractor.rb:48`) and `ScheduledJobExtractor` (`scheduled_job_extractor.rb:96`) call `YAML.safe_load` without `aliases: true`. On Psych 4+ (Ruby 3.1+, the gem's supported range) any file using YAML anchors/aliases raises `Psych::AliasesNotEnabled`, which the surrounding `rescue StandardError` converts into a logged skip — the whole file vanishes from the index with no unit and no `woods:validate` signal.

`<<: *defaults` is a widespread idiom in exactly these files: locale files sharing keys across languages, and environment-nested `config/recurring.yml` with a `default: &default` anchor. Reproduced mechanically.

## Suggested fix

Pass `aliases: true` to both `safe_load` calls — alias expansion of the already-permitted scalar/hash types adds no object-deserialization risk. For `recurring.yml`, consider also stripping/evaluating ERB before parsing, and rescue `Psych::SyntaxError` into a warn that names the reason. Specs with anchored YAML for both extractors.

Tracked as **B-091** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #204: Whenever block parser truncates at the substring 'end' inside class/command names
## Description

`parse_whenever_blocks` (`lib/woods/extractors/scheduled_job_extractor.rb:196`) uses `/every\s+(.+?)\s+do\s*\n(.*?)end/m` — the terminating `end` is neither word-bounded nor line-anchored, so the lazy body match stops at the first literal `end` **anywhere**, including inside identifiers like `CalendarSyncJob`, `WeekendDigest`, or `RecommendationJob` (also `append`, `send`).

Reproduced with the real methods: `every 1.day do\n  runner "CalendarSyncJob.perform_later"\nend` captures body `  runner "Cal` — the truncated body fails `detect_whenever_command`'s quoted-string regexes, yielding `command_type :unknown`, no `job_class`, and no `:job` dependency edge. The schedule→job link the extractor exists to capture is silently lost.

## Suggested fix

Anchor the terminator to a line — `/every\s+(.+?)\s+do\s*\n(.*?)^\s*end\s*$/m` — or reuse the line-by-line depth tracker already present in `RakeTaskExtractor#extract_task_block`. Spec with a job name containing "end".

Tracked as **B-092** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #205: PageRank drops rank mass for duplicate edges and edges to unregistered targets
## Description

PageRank (`lib/woods/dependency_graph.rb:294`) counts out-degree over **raw** edges (`(@edges[src] || []).size`) but sums incoming contributions over the deduplicated `@reverse` Set — one contribution of `scores[src]/out_degree` per unique source. A source with two edges to the same target — documented as normal: the same target referenced via `:belongs_to` **and** `:code_reference` (see `#dependents_detail`'s own doc, :239-244) — delivers only half its intended mass; the rest vanishes. Likewise a node whose edges point at unregistered targets leaks that mass entirely yet is not counted in `dangling_sum` (which requires edges to be empty, :285).

Reproduced mechanically: total mass decayed to **0.46 vs 1.0**; the duplicate-edge target scored 0.271 vs the 0.649 control. Net effect: systematic under-ranking of exactly the most-referenced units (the belongs_to + constant-mention shape is the mainline case), feeding retrieval ranking and the pagerank exposed via MCP/Obsidian.

The existing sum-to-one spec passes because its fixtures have neither duplicate edges nor unresolvable targets.

## Suggested fix

Weight each source's contribution by its edge multiplicity to the node (precompute a target→count map per source, keeping it O(edges)); compute out-degree over edges whose target is registered and treat zero-resolvable nodes as dangling. Extend the sum≈1.0 spec with a duplicate-edge and an unresolved-target case.

Tracked as **B-093** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #206: SnapshotStore re-capture at the same git SHA orphans every previous unit-hash row (unbounded SQLite growth)
## Description

`capture` (`lib/woods/temporal/snapshot_store.rb:233-238`) upserts the snapshot row with SQLite `INSERT OR REPLACE` keyed on `UNIQUE(git_sha)`. REPLACE deletes the old row and inserts a new one with a **new** AUTOINCREMENT id; the subsequent `DELETE FROM woods_snapshot_units WHERE snapshot_id = ?` uses the new id (fetched after the replace) and deletes nothing. `PRAGMA foreign_keys` is never enabled anywhere, so nothing errors: the old snapshot's full unit-hash set stays behind keyed to a snapshot id that no longer exists.

Reproduced mechanically. With `enable_snapshots` on, the ordinary edit-extract loop at an unchanged HEAD (uncommitted work — the normal development state) leaks one whole unit set per extraction: ~7,000 orphaned rows per run on a production-shaped host, unboundedly, degrading the `idx_snapshot_units_identifier` index that `unit_history` scans, with no query able to reach or clean the dead rows.

## Suggested fix

Preserve the rowid: `INSERT ... ON CONFLICT(git_sha) DO UPDATE` (SQLite ≥ 3.24) or fetch the existing id first and UPDATE — then the DELETE actually clears the prior unit set. Regression spec: `COUNT(*) FROM woods_snapshot_units == units.size` after a double capture at one SHA.

Tracked as **B-094** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #207: Notion exporter has no change detection — every sync re-upserts every page (2 API calls each at 3 rps)
## Description

Unlike the Unblocked exporter (versioned SyncManifest skips unchanged documents) and the Obsidian exporter (byte-identical idempotent writes), the Notion exporter has no manifest, hash, or skip path: `sync_all` unconditionally upserts every model page and every column page on every run, and each upsert costs two API calls — a `find_page_by_title` databases/query POST plus create/update (`lib/woods/notion/exporter.rb:186-196`) — serialized through the 3 req/sec RateLimiter. `@page_id_cache` only links columns to parents; it never skips the per-upsert title query.

Cost model: a 300-model app averaging 20 columns/model = ~6,300 pages = ~12,600 API calls ≈ **70 minutes wall-clock per `woods:notion_sync`**, including fully no-op runs. A nightly schedule pays that daily against an unchanged schema.

## Suggested fix

Reuse the Unblocked pattern: persist a Notion sync manifest (title/page_id + content hash of mapped properties) under index_dir, skip unchanged pages, and only query Notion when the manifest lacks a page_id — which also removes ~half the calls for changed pages. Fix #149 (title keying) first so manifest keys are correct.

Tracked as **B-095** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #208: write_unit_file byte-compare skip is inert across runs: extracted_at defeats it
## Description

The documented fsync-avoidance optimization — skip the atomic write when the serialized unit is byte-identical to disk (`lib/woods/extractor.rb:919-924`) — can essentially never fire in its motivating scenario, because `ExtractedUnit#to_h` embeds a fresh wall-clock timestamp (`extracted_at: Time.now.iso8601`, `extracted_unit.rb:59`) in every serialization. The payload always differs from the previous run's file in that one field, so `identical_on_disk?` returns false and the write proceeds.

Net effect on the motivating case (routes change → wholesale replacement of ~1,707 route-consumer units, per CLAUDE.md): the optimization **adds** a full read of every file and then writes anyway — strictly worse than no optimization. (Within a single run it still guards against double-writes, which is why nothing observably breaks.)

## Suggested fix

Make the comparison timestamp-insensitive: compare with the volatile key masked (mirroring `VOLATILE_UNIT_KEYS` in `spec/support/index_comparison.rb`), keeping the older timestamp on disk — the equivalence oracle already masks it. Or move `extracted_at` out of per-unit JSON entirely; it is per-run data that belongs in the manifest. Also freeze time in the covering spec so it cannot flake.

Tracked as **B-096** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review; adversarially verified.

### LINKED ISSUE #209: MetadataStore::SQLite#search interpolates field names into SQL unparameterized (latent injection + unescaped LIKE metacharacters)
## Description

`search(query, fields:)` (`lib/woods/storage/metadata_store.rb:277`) builds `json_extract(data, '$.#{'{'}field{'}'}') LIKE ?` by direct string interpolation of each field name. A field value like `x') OR 1=1 OR json_extract(data,'$.x` breaks out of the JSON-path literal, injects a tautology, preserves the bind-param count, and returns all rows — confirmed in a standalone sqlite3 script.

**Latent, not live:** production retrieval callers pass no `fields:` today (`search_executor.rb:175,324`). But the Interface documents `fields` as a public parameter, and the pgvector adapter's `build_where` shows the house standard (regex-validate + quote keys, `pgvector.rb:197-218`). Separately, `%`/`_` metacharacters from the query are interpolated unescaped into the LIKE pattern in **both** metadata adapters, silently broadening matches.

## Suggested fix

Validate field names against a whitelist regex (mirroring `build_where`) before interpolation, and escape LIKE metacharacters in the query (`ESCAPE` clause). Low priority — hardening before any caller grows a `fields:` argument.

Tracked as **B-097** in `docs/backlog.json`. Found by the 2026-07-30 full-gem review (verified as plausible/latent).

LeahArmstrong:
## Summary

A full adversarially-verified review of the gem (12 subsystem reviewers + independent verification of every medium+ finding) produced 27 new issues (#183–#209) and confirmed every substantive pre-existing one. This PR fixes **all of them**: 30 tracked defects across 26 fix commits, each with regression specs proven to fail against the reverted code. The only review outputs *not* addressed here are the feature requests #10/#12 (Stimulus, seed-data extraction); #11/#83 were found already satisfied and are closed.

A post-hoc claims audit re-verified every fix claim below against the final tree (~40 spec files, ~1,800 examples re-run): all 30 substantiated, none false or exaggerated.

## Motivation

Closes the following (one keyword per issue so merge closes them all):

Fixes #149. Fixes #169. Fixes #170. Fixes #174. Fixes #175. Fixes #176. Fixes #177. Fixes #178. Fixes #179.
Fixes #183. Fixes #184. Fixes #185. Fixes #186. Fixes #187. Fixes #188. Fixes #189. Fixes #190. Fixes #191. Fixes #192. Fixes #193. Fixes #194. Fixes #195. Fixes #196. Fixes #197. Fixes #198. Fixes #199. Fixes #200. Fixes #201. Fixes #202. Fixes #203. Fixes #204. Fixes #205. Fixes #206. Fixes #207. Fixes #208. Fixes #209.

**Addresses (deliberately not closing):** #150 — items 2 and 5 fixed here; item 1's wrong-result race was already fixed by the refcounted-pin work, with the benign LRU-drift residual tracked in #218. #181 — the adapter-robustness half is fixed here; the root cause (B-062 identifier collisions) is the design proposal in #213.

**Follow-ups filed for the next agent:** #211 (durable-backend vector pruning), #212 (evaluation harness), #213 (B-062 root-fix design), #214 (IndexValidator), #215–#218 (grouped polish batches), #219 (test hygiene), #220 (CI gaps) — mirrored as B-098–B-107 in `docs/backlog.json`.

The full review report (methodology, verifier notes, subsystem health): https://claude.ai/code/artifact/807639a4-be3e-4231-9031-d25f972064fc

## Changes

**Host-app safety (#183)** — console-MCP guards are path-scoped and enablement is decided at request time; enabling the documented feature no longer 401s the whole host app, and flags set in `config/initializers/woods.rb` now work (previously a silent no-op — railtie initializers run before config initializers). Disabled console path now passes through to the app instead of answering 410 (docs updated; the v1.2.0 security-validation record is annotated, not rewritten).

**Retrieval correctness (#184–#186, #150 item 5)** — type-filtered `codebase_retrieve` returned empty on every booted server (string/symbol key mismatch in the dump backfill, on boot *and* reload); the ranking layer was largely decorative (dead `matched_fields` signal, raw RRF swamped by 0–1 metadata weights, assembler re-sorting away ranked order, PageRank memo never invalidated on reload); classifier patterns hard-filtered vector search on words like "get" and "email" with no fallback; framework units were formatted into two context sections at once. All fixed; vector-store filter matching is now key-form-indifferent.

**Extraction fidelity (#174–#177, #193, #194, #199–#204)** — concern inlining works for compact-style declarations and CallbackAnalyzer finally sees concern-defined callbacks (models *and* controllers — controllers also gain `:concern` edges and inlining parity); one shared `SourceNesting` parser fixes namespace derivation in five source-parsing extractors; ApplicationController is indexed for the first time; polymorphic associations get `via: :polymorphic_interface` instead of phantom model edges; per-call cache attribution; GraphQL outer identifiers; YAML `aliases: true`; line-anchored Whenever terminator; label-form rake tasks + string namespaces; `RSpec.describe Klass, type: :model` test mapping; full extraction sweeps orphaned unit files (they were being resurrected into `_index.json` by the next incremental run).

**Pipeline integrity (#169, #170, #198, #208)** — `rails_source`/`gem_source` become a Gemfile.lock-keyed whole-app extractor through the real write pipeline (the old `extract_framework` bypassed AtomicFile, `_index.json`, manifest, lock, and generation; it's now a locked alias), with the previously-dead `include_framework_sources` knob genuinely gating participation; every index writer — `woods:clean` (daemon-aware, lock-file-sparing), both embed tasks, and MCP `pipeline_extract` — now takes the pipeline lock; failed extractor construction and degraded eager-load boots can no longer license unit pruning; the write-skip byte-compare actually fires now (`extracted_at` masked).

**Embedding durability (#187, #188, #191, #192, #181-adapter, #205, #206)** — the resilience layer is wired for the first time (a single 429 aborted embed runs and discarded paid-for work; providers now retry with Retry-After honored, per-instance breakers); a mis-pointed incremental embed can no longer wipe the vector index (empty-load refusal + 30% purge guard); WVF1 ids hydrate as UTF-8 (non-ASCII identifiers re-embedded forever and crashed the dump rewrite); pgvector works via the documented setup (the generator created schema nothing used; `ensure_schema!` had zero callers) and dedupes in-batch ids naming the dropped ones; PageRank keeps rank mass for duplicate/unresolvable edges; SnapshotStore stops leaking a full unit set per same-SHA re-capture.

**Robustness and exports (#149, #150 item 2, #189, #190, #195–#197, #207, #209)** — `AtomicFile` discipline extended to the embed pipeline, Unblocked manifest, StatusReporter, and flow layer (three confirmed `LANG=C` crashers); flow artifacts are relative + atomic with a reassembly fallback; the watch daemon's stand-down no longer clobbers the live daemon's status and the lock heartbeat can't resurrect a released lock; chunkers track nested control flow and endless/operator defs; Notion multi-model sync no longer corrupts the Columns DB (qualified `table.column` titles + relation-verified legacy adoption) and gains a sync manifest (unchanged pages = zero API calls); both API clients stop retrying non-idempotent POSTs on read timeout; metadata search validates field names and escapes LIKE metacharacters.

**Docs (#179 + drift batch)** — eight verified drift spots corrected against code: the WATCH_DAEMON lock-wait row, the real `WOODS_MCP_HTTP_ALLOWED_ORIGINS` var, the post-B-053 `console_eval` gating, the embedded Tier-4 opt-ins, a working Qdrant config example, the two missing MCP tools, the incremental contract's post-prune reconciliation, and COVERAGE_GAP_ANALYSIS's backlog pointers.

**New capability (#178)** — `embedding_provider = :fake` (deterministic, offline, promoted from spec support with MCP round-trip wiring) plus provider-object injection; `woods:retrieve` resolves all four backends through configuration instead of hardcoding Ollama + in-memory stores. The previously-impossible offline embed→retrieve round trip is now an integration spec.

Backlog: B-060–B-061, B-067–B-068, B-071–B-097 marked resolved; B-098–B-107 added for the follow-ups, all cross-referenced to issues.

## Testing

- Full suite after every batch; final state: **6,030 examples, 0 failures, 2 pending** (the pre-existing tiktoken pendings) — up from 5,482 before the fixes (~550 examples added). RuboCop: **533 files, no offenses** (plus the one chain-alignment offense CI's rubocop flagged, fixed in `7e7e103`).
- Every fix carries regression specs **verified to fail against the temporarily reverted code**. Several were reproduced mechanically in standalone Ruby before fixing (PageRank mass decay to 0.46, the daemon status clobber, the Notion collapse, the `LANG=C` crashers); the railtie fixes were smoke-validated against a real booted railties 8.1 app.
- Post-hoc verification: a claims audit re-confirmed all 30 fixes with file:line evidence and passing specs, and a seam review covered the files multiple fix batches touched.
- CI: full matrix green on the first run (all booted-extraction rows Rails 6.0→8.0, Ruby 3.0→4.0) apart from the one lint offense, fixed.
- **Reviewer notes:** (1) the `:booted_app` incremental-equivalence harness runs in CI's rails-matrix (couldn't run in this container) — worth confirming green since `extractor.rb` changed on the incremental path; (2) `spec/extractors/controller_extractor_fixture_spec.rb` has a **pre-existing** standalone-only failure (tracked in #219); (3) upgrade note for hosts that ever ran the old `woods:extract_framework`: one `woods:clean` + `woods:extract` clears the old-naming orphan files, and the new `woods:validate` warnings surface the condition; (4) identifiers change for previously-misnamed namespaced units (#174) — that is the fix; a full extraction rewrites them.

## Checklist

- [x] Tests added/updated
- [x] `bundle exec rake spec` passes (6,030 examples, 0 failures; run via the rspec exe path — the bundler binstub is broken in this container)
- [x] `bundle exec rubocop` passes (533 files, no offenses)
- [x] CHANGELOG.md updated (if user-facing change)
- [x] Documentation updated (if applicable) — CONSOLE_MCP_SETUP, CONFIGURATION_REFERENCE, EMBEDDING_MODELS, GETTING_STARTED, USE_CASES_AND_FEATURE_GAPS, WATCH_DAEMON, TROUBLESHOOTING, BACKEND_MATRIX, MCP_SERVERS, INCREMENTAL_EXTRACTION, COVERAGE_GAP_ANALYSIS, security plan annotation, CLAUDE.md counts

https://claude.ai/code/session_01MNAhw97QTooKmFEUwcSZpo


LeahArmstrong:
## Merge-readiness verification

Two independent verification passes ran against this branch before merge:

**Claims audit** — every `Fixes` claim re-verified against the final tree with `file:line` evidence and its regression spec re-run (~40 spec files, ~1,800 examples): **all 30 fix claims substantiated, none false or exaggerated**. The audit's three mechanics flags are resolved: the PR body now carries one closing keyword per issue (36 issues close on merge; #150/#181 deliberately stay open with residuals tracked in #218/#213), #179 was added to the closing list, and CLAUDE.md's stale "four writers" lock roster was corrected.

**Seam review** — a full-diff review focused on files touched by multiple fix batches. It cleared the high-risk interactions (extractor sweep vs write-skip vs rails_source gating; resilience wrap order; the retrieval signal chain end-to-end; Notion manifest vs #149 titles; lock lifecycles; no process debris, no weakened assertions) and surfaced **three real findings, all fixed in `e4bccd4`**:
- injected provider objects (the #178 feature) crashed MCP boot at the probe — now duck-typed as reachable;
- MCP `pipeline_embed` was a seventh unlocked index writer — now locked like its rake twins and `pipeline_extract`;
- the GraphQL identifier fix hand-rolled an end-blind nesting scan — now on the shared `SourceNesting` parser.

One pre-existing (on `main`) finding was filed instead of fixed: ranker signals go neutral for chunk-suffixed candidates → #218.

**Final state**: 6,041 examples, 0 failures, 2 pending (pre-existing tiktoken); rubocop 533 files, no offenses. CI is running on this head; every prior run was fully green apart from the since-fixed lint offense — the booted-extraction matrix (Rails 6.0→8.0, Ruby 3.0→4.0) passed on the first attempt, which covers the incremental-path changes this container couldn't validate locally.

**Hand-off for the next agent**: follow-ups are #211–#220 (mirrored as B-098–B-107 in `docs/backlog.json`), plus feature requests #10/#12. #11/#83 are closed as satisfied.

---

